PR #2, opened 2018-04-25
Jhe java8mig178

apgsga-jhe:
Pull request for java8mig178 and java8mig-324.
The command line tool has 2 new option:

- vv -> to validate all module for a given version
- vp -> to validate all modules within a path

One can not save a patch if a module name contained within the patch is invalid.

Before updating/installing the patch-server, one still needs to increase its version number, this can be done after having merge with the head.

### DIFF
--- a/apg-patch-service-server/packaging/rpm/pre-install.sh
+++ b/apg-patch-service-server/packaging/rpm/pre-install.sh
@@ -14,5 +14,7 @@ if [ "$1" = "1" ]; then
 	/usr/sbin/useradd -r -m -c "apg-patch-service-server user" apg-patch-service-server -g apg-patch-service-server 2> /dev/null || :
 	echo "Adding user apg-patch-service-server to *domain users*"
 	/usr/sbin/usermod -G domain\ users apg-patch-service-server 2> /dev/null || :
+	setfacl -R --modify g:apg-patch-service-server:rwX /var/local/cvs/root
+	setfacl -R --default --modify g:"domain users":rwX,g:jenkins:rwX,g:apg-patch-service-server:rwX /var/local/cvs/root
 fi
 exit 0
--- a/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchOpServiceController.java
+++ b/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchOpServiceController.java
@@ -2,6 +2,7 @@ package com.apgsga.microservice.patch.server;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang.NotImplementedException;
 import org.apache.commons.logging.Log;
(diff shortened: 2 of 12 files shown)


apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/SimplePatchContainerBean.java
```diff
@@ -108,13 +116,20 @@ private void preProcessSave(Patch patch) {
 			createBranchForDbModules(patch);
 			jenkinsClient.createPatchPipelines(patch);
 		}
+		System.out.println(patch.getMavenArtifacts());
```
chhex: This should go away or use the Standard Logger ;-)

apg-patch-service-server/packaging/rpm/pre-install.sh
```diff
@@ -14,5 +14,7 @@ if [ "$1" = "1" ]; then
 	/usr/sbin/useradd -r -m -c "apg-patch-service-server user" apg-patch-service-server -g apg-patch-service-server 2> /dev/null || :
 	echo "Adding user apg-patch-service-server to *domain users*"
 	/usr/sbin/usermod -G domain\ users apg-patch-service-server 2> /dev/null || :
+	setfacl -R --modify g:apg-patch-service-server:rwX /var/local/cvs/root
```
chhex: I would maybe comment the reason why this is nessary or refer to the Jira tick

apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchOpServiceController.java
```diff
@@ -183,5 +186,19 @@ public void clean() {
 	public void init() throws IOException {
 		throw new UnsupportedOperationException();
 	}
+	
+	@RequestMapping(value = "/validateArtifactNamesFromVersion", method = RequestMethod.GET)
```
chhex: Do not quite understand why in one case we have a POST und in the other a GET Request, where as both Method return a Map of MavenArtifacts, but no Response Body?

apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/SimplePatchContainerBean.java
```diff
@@ -108,13 +116,20 @@ private void preProcessSave(Patch patch) {
...
-				.forEach(art -> addModuleName(art));
+				.forEach(art -> addModuleName(art,patch.getMicroServiceBranch()));
 	}
 
-	private MavenArtifact addModuleName(MavenArtifact art) {
+	private MavenArtifact addModuleName(MavenArtifact art, String cvsBranch) {
 		try {
 			String artifactName = am.getArtifactName(art.getGroupId(), art.getArtifactId(), art.getVersion());
+			
+			Map<String, List<MavenArtifact>> wrongNames = getArtifactNameErrorAsMap(Lists.newArrayList(art),cvsBranch);
```
chhex: Do not understand this just quite: Basically you just need to check the artifactName against the CVS, so it's a check of a single Module?

apgsga-jhe:
No sure I get what you mean, I don't believe I added an endpoint to validate a single artefact, or did I ?!?!

chhex:
As discussed

PR #6, opened 2018-05-08
Che java8 mig 340

chhex:
Tentativer Pull Request, vorbehältlich lokalen Integrationstests


apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/persistence/AtomicFileWriteManager.java
```diff
@@ -0,0 +1,88 @@
...
+ * 8. 7 is the critical path: if 7 fails , we may have a corrupt state of the
+ * Target File, but we have a Backup in the Backup Directory
+ * 
+ * 
+ * As a temporary solution, we use the dormant Apache commons library:
+ * https://commons.apache.org/proper/commons-transaction/ Specifically:
+ * https://commons.apache.org/proper/commons-transaction/file/index.html Which
+ * is "dormant", but which seems to working just fine. Better, then probably ,
+ * if i would write my own code and in much less time ;-)
+ *
```
apgsga-jhe: Quick question: I didn't go through the source code of _org.apache.commons.transaction.file.FileResourceManager_ , but if I understood it well, all the above steps are managed by the _org.apache.commons.transaction.file.FileResourceManager_ . org.apache.commons.transaction.file.FileResourceManager basically works with the "workDir" and the "tempDir" ... and if something goes wrong, it will "automatically" put the file from "tempDir" back to "workDir". Is that correct?

apgsga-jhe:
Looking good to me, I just ask one question along the code :) (gave me as well the possibility to test Github code review...)

apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/persistence/AtomicFileWriteManager.java
chhex: Actually there isn't any source in the Repo, only in Apache's subversion repo ;-( 
The comments are maybe somewhat misleading. They actually  adress what i would intend, if i would implement  it (TODO!?), 
The commons.transaction Code has two directories a workDir and a targetDir.  Here a somewhat short version: It does all the preparing work in workDir and the moves the files over to the TargetDir, when respectively on commit. It better explained here: 
https://commons.apache.org/proper/commons-transaction/apidocs/index.html
I ll will add this link to the code and make it clearer that the discription is "intended"

chhex: Shall i maybe delete the comments, since there are not relevant currently?

apgsga-jhe: Yes maybe, might be easier when someone new will have a look at it in the future.

chhex: Yep agreed.  I deleted the description. You should be able to see the changes

apgsga-jhe: Yep I'm seeing it.

PR #11, opened 2018-05-23
First Commit for Exception Handling For Review only (NOT TO BE MERGED)

chhex:
Discuss the general approach:

1. Specific Error resp. Exception Messagre Object in /apg-patch-service-common com.apgsga.microservice.patch.api.PatchErrorMessage
2. Each Runtime Exception in /apg-patch-service-server get a Class specific Exception Class in com.apgsga.microservice.patch.exceptions
3. Each throw new RuntimeException  in /apg-patch-service-server  gets replaced by one of above (2.)
4. For Asserts in the Code per standard org.springframework.util.Assert is used
5. The Exception Messages for 3 and 4 are made more explicit to the situation
6.  Spring Controller Advice Class : com.apgsga.microservice.patch.server.PatchServiceExceptionHandler
With @ExceptionHandler 's for each of 2 and 4 
7. The clients apscli and the Java Client need to have a specific Error Handler for the Resttemplate 
The one used within the Itui needs to comply to JavaForms Exception Handlling and the Apicli basically one that simply sets the Exit Code to Error with a println of the Error Message

Don't know if it's a overkill, specifically the specific Exception per Class.  But hen again it helps to locate to Exception

### DIFF
--- a/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchServiceExceptionHandler.java
+++ b/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchServiceExceptionHandler.java
@@ -1,0 +1,53 @@
+package com.apgsga.microservice.patch.server;
+
+import java.util.Optional;
+
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+import com.apgsga.microservice.patch.api.PatchErrorMessage;
+import com.apgsga.microservice.patch.exceptions.AtomicFileWriteManagerException;
+import com.apgsga.microservice.patch.exceptions.GroovyScriptActionExecutorException;
+import com.apgsga.microservice.patch.exceptions.JschExecutionException;
+
+public class PatchServiceExceptionHandler extends ResponseEntityExceptionHandler {
+
+	protected final Log LOGGER = LogFactory.getLog(getClass());
+
+	@ExceptionHandler(JschExecutionException.class)
+	public ResponseEntity<PatchErrorMessage> notFoundException(final JschExecutionException e) {
+		return error(e, HttpStatus.INTERNAL_SERVER_ERROR);
+	}
+
+	@ExceptionHandler(GroovyScriptActionExecutorException.class)
+	public ResponseEntity<PatchErrorMessage> notFoundException(final GroovyScriptActionExecutorException e) {
+		return error(e, HttpStatus.INTERNAL_SERVER_ERROR);
+	}
+
+	@ExceptionHandler(AtomicFileWriteManagerException.class)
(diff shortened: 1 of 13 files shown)


apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchServiceExceptionHandler.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	// TODO (che,23.5) : First go, to be discussed
+	private ResponseEntity<PatchErrorMessage> error(final Exception exception, final HttpStatus httpStatus) {
+		final String errorText = Optional.of(exception.getMessage()).orElse(exception.getClass().getSimpleName());
+		Throwable cause = exception.getCause();
+		final String causeMsg = cause != null ? cause.getMessage() : "<Root Cause>";
+		PatchErrorMessage errorMsg = new PatchErrorMessage(errorText, causeMsg,
+				ExceptionUtils.getFullStackTrace(exception));
+		LOGGER.warn(errorMsg.toString());
```
apgsga-jhe: Don't we want to log as error instead of warning?

chhex: Agree, we could. The ratio: the are actually caught and handled exception, so they get to client in an "orderly" way, as PatchErrorMessage, with corresponding Http Codes. The client can then decide, what he wants to do with the message. 

Vs error: Unhandled Exceptions, which are propagated to the client as exceptions

What do you think?

chhex:
@apgsga-jhe Concerning the overkill :  I was just thinking to reduce all the specific RuntimeExceptions Classes - per Class with Runtime Exceptions, to one. Basically all the classes do the same. So we would have one Exception class, instead of how many? Five, six, which do all the same thing.  What we could do instead is localize the Error messages, so they would be in a resource file and identified by a id. This way traceability is also guaranteed.
And we have less code, which is always a good thing

apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchServiceExceptionHandler.java
apgsga-jhe: Yep sure, definitely much better than Unhandled Exception getting back to the client. Ok make sense, I think we just have to be careful not to log error as warning on the server side even if the information is returned to the client. In case of a real error (I mean, not the client asking for something which would not be valid), might be "difficult" to determine what was more or less important. However as you said, here it is only just before returning to the client, meaning we would eventually have a server error logged somewhere else before. Then yep, I agree.

apgsga-jhe:
@chhex : yep good point, make sense -> code reduced and still easy to identify where things happenened. Agreed.

apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/PatchServiceExceptionHandler.java
chhex: We could maybe do something like: 
LOGGER.warn("Returning Handled Exception to Client " + errMsg.toString)

apgsga-jhe: 100% agreed.

chhex:
@apgsga-jhe so i am closing the pull request ( since it never was one, haha ;-)

PR #12, opened 2018-05-28
JAVA8MIG-353: support onClone call on cmd line, part "c" of JAVA8MIG-353

apgsga-jhe:
This pull request if for point "c" of JAVA8MIG-353 (see CHE comment of 24.05.2018). I would say this is not 100% done yet, as couple of variable value should be fetch (I believe) from a configuration file, or at least should be better centralized. But apart from this, it seems to be working fine.


apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+	
+	private final String RELEASE_REPO = "releases"
+	
+	private Artifactory artifactory;
+	
+	private revisionFilePath
+	
+	public PatchCloneClient(String p_revisionFilePath) {
+		// TODO JHE: Get these from a configuration file. Maybe a new file: /var/opt/apg-patch-common/artifactory.properties ?? 
+		def url = "https://artifactory4t4apgsga.jfrog.io/artifactory4t4apgsga"
```
chhex: I would actually have these properties private to apscli , that is in /etc/opt/apg-patch-cli, see also latest arguments https://jira.apgsga.ch/browse/JAVA8MIG-363

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+		if(lastRevisionForTarget == null) {
+			throw new RuntimeException("No revision found for ${target}.")
+		}		
+		
+		return lastRevisionForTarget	
+	}
+	
+	private void deleteRevisionWithinRange(Long lastRevision) {
+		
+		// TODO JHE: Ideally the range step should be centralized somewhere
```
chhex: Agreed , thinking of it. Maybe it would be even better , if we move all the revision to here and call apscli in all szenarios, also from the Patch Pipeline

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List
+import org.jfrog.artifactory.client.Artifactory
+import org.jfrog.artifactory.client.ArtifactoryClientBuilder
+import org.jfrog.artifactory.client.model.RepoPath
+import groovy.json.JsonBuilder
+import groovy.json.JsonSlurper
+
+class PatchCloneClient {
+	
+	private final String PROD_TARGET = "CHPI211"
```
chhex: This needs to come via /var/opt/apg-patch-common/TargetSystemMappings.json

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+			removeArtifacts(RELEASE_REPO, "*${FIRST_PART_FOR_ARTIFACT_SEARCH}${from}*", false)
+			from++
+		}
+	}
+	
+	private void resetLastRevision(String target) {
+		
+		def revisions = getParsedRevisionFile()
+		
+		// TODO JHE: Is this part really correct? I mean that we eventually set it to SNAPSHOT? Not sure it will even happen, but ...
```
chhex: Agreed , but i think it's correct: I means  the Target hasn't been patched.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+		deleteRevisionWithinRange(lastRevision)
+		resetLastRevision(target)
+	}
+	
+	private Long getLastRevisionForTarget(String target) {
+		
+		def revisions = getParsedRevisionFile()
+		
+		def lastRevisionForTarget = revisions.lastRevisions[target]
+		if(lastRevisionForTarget == null) {
```
chhex: I don't think we need to throw here an exception? A possible szenario? That means the clone before the first Patch for the Target.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
chhex: I think releases should be fine. Or at least that's the way we want releases configured in Artifactory.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+		else {
+			throw new RuntimeException("Error while parsing Revision file. ${revisionFile} not found.")
+		}
+		
+		return revisions
+	}
+	
+	private void listArtifacts(String repoId, String regex) {
+		List<RepoPath> searchItems = artifactory.searches().repositories(repoId).artifactsByName(regex).doSearch();
+		System.out.println("Listing all Artifacts for repo " + repoId);
```
chhex: Would use println -> small advantages of Groovy ;-)

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+		searchItems.each{repoPath -> 
+			System.out.println(repoPath.getItemPath());
+		};
+		System.out.println("Total number of Artifacts in " + repoId + ": " + searchItems.size());
+		
+	}
+	
+	private void removeArtifacts(String repo, String regex, boolean dryRun) {
+		if(dryRun) {
```
chhex: Would do the dryRun switch below

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -0,0 +1,114 @@
...
+			println "Dry run only ... following Artifacts would have been deleted";
+			listArtifacts(repo,regex);
+		}
+		else {
+			List<RepoPath> searchItems = artifactory.searches().repositories(repo).artifactsByName(regex).doSearch();
+			searchItems.each{repoPath ->
+				artifactory.repository(repo).delete(repoPath.getItemPath());
```
chhex: Would do the dryRun switch here on the delete statement. And log would be / is deleted

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: Yep good point, make sense. I'll comment the code accordingly, and link java8mig-363 to java8mig-353.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: yep

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: Correct, I forgot that this file actually contains what is required. will do it.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: "That means the clone before the first Patch for the Target." : yep that's what I thought. What would happen if we clone a target for which we never had any patch installed on? Mmhh, ok, actually we could just say that if "lastRevisionsForTarget" is null, then we just don't do anything (no delete, no reset ...), seems to be as simple as that, right?

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: yep sure, stupid copy/paste ... :)

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: Ok good point, then I can actually delete "listArtifacts" methods.

apgsga-jhe:
I'll do the adjustments as written, and request a new review then ...

chhex:
The question is also if, and when yes when we want to move the revision logic into apscli. And change the contract in the Pipeline? Shall open a new ticket?

apgsga-jhe:
I was actually writing for this ;) . 
I believe we should open a new ticket to at least discuss it, maybe after a second thought. For now I see at least one big advantage: the path to the Revision file would then only be known by the patchCli .. and therefore configured at one private place. I would need to have a look back in the pipeline to evaluate what would change there, but I don't believe that much (probably 2 functions ...)

PR #13, opened 2018-06-01
Java8mig365 centralize revision logic

apgsga-jhe:
Revision.json has been moved into /var/opt/apg-patch-cli folder. RetrieveRevision and SaveRevision have now been moved into apscli.
Careful: this pull request is only to manage the Revision from apscli. The clone problem (JAVA8MIG-353) still have to be done correctly, and will be done as soon as this one will have been approved and merged on master.

### DIFF
--- a/apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
+++ b/apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
@@ -13,6 +13,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 
 import groovy.json.JsonSlurper
 import groovy.json.JsonBuilder
+import groovy.json.JsonOutput
 
 class PatchCli {
 
@@ -337,6 +338,18 @@ class PatchCli {
 				error = true
 			}
 		} 
+		if (options.rr) {
+			if(options.rrs.size() != 3) {
+				println "3 parameters are required for the retrieveRevision command."
+				error = true
+			}
+		}
+		if (options.sr) {
+			if(options.srs.size() != 3) {
+				println "3 parameters are required for the saveRevision command."
+				error = true
+			}
+		}
 		if (error) {
 			cli.usage()
 			return null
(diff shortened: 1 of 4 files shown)


apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/testRevision.groovy
```diff
@@ -0,0 +1,280 @@
+package com.apgsga.patch.service.client
+import groovy.json.JsonBuilder
+import groovy.json.JsonSlurper
+
+revisionFileName = "src/test/resources/Revisions.json"
+
```
chhex: Not important, but still: Generally i don't like commented out  code ;-) ...  either it's code, which is needed, or not. If it is history fine, but the VCS does tracking of history better ;-)

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -550,7 +639,7 @@ class PatchCli {
 		}
 		// TODO JHE: get the path to Revision file from a configuration file, or via parameter on command line?
 		// 			 will/should be improved as soon as JAVA8MIG-363 will be done.
-		def cloneClient = new PatchCloneClient("/var/jenkins/userContent/PatchPipeline/data/Revisions.json","/var/opt/apg-patch-common/TargetSystemMappings.json")
+		def cloneClient = new PatchCloneClient(revisionFilePath,"${configDir}/TargetSystemMappings.json")
```
chhex: Just a comment: Yes, we need to see, how we configure the different usage scenarios, best.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -536,10 +563,72 @@ class PatchCli {
 		println "Performing onClone for ${options.ocs[0]}"
 		// TODO JHE: get the path to Revision file from a configuration file, or via parameter on command line?
 		// 			 will/should be improved as soon as JAVA8MIG-363 will be done.
```
chhex: Exactly, yes also my general comments.

PR #17, opened 2018-06-12
Package dependency and seperated ops and application properties file

chhex:


chhex:
Pull request for merge after Thuesday afternoon. @apgsga-jhe  can merge and install, when it is convenient.

chhex:
Added in a first iteration rpm deployment to a Artifactory rpm repository.  Tested it locally with a Centos Image  with Yum. could be basis to discuss with Ueli and Patric.

apg-patch-service-cmdclient/build.gradle
```diff
@@ -116,8 +116,10 @@ ospackage {
 	os = LINUX
 	type = BINARY
 	arch = NOARCH
+	requires('aps' , "${savedVersion}",  EQUAL)
 	
-	postInstall file("packaging/rpm/post-install.sh")
+	// Since Package while be used from Jenkins pipeline for writeing and reading revision file
+	postInstall 'chown -R jenkins:jenkins /var/opt/apg-patch-cli'
```
apgsga-jhe: not sure here if it will be fine if /var/opt/apg-patch-cli doesn't exist. I don't see the "directory" command which should eventually create it. I'm still going ahead and will merge, I'll fix it on master if it causes any problem.

chhex: Good point , i optimized away the directory statement .... absolutely needs to be there. I fix that quickly

PR #18, opened 2018-06-14
Rpm yum repo deployment 

chhex:
Rpm yum repo deployment first go for request for comments
Note: contains also the changes of Pull request https://github.com/apgsga-it/piper/pull/17
As soon as 17 is merged they should go away, any way i ll update then from upstream.


chhex:
yumrep is defined as RPM repository in our artefactory : https://artifactory4t4apgsga.jfrog.io/artifactory4t4apgsga/webapp/#/admin/repositories/local
Since we are using the 'com.jfrog.artefactor' Plugin for publishing , that seems enough

The "important" part is that this repository can be used resp. configured as YUM repo on the centos platforms of our patch services. 
We need to discuss (tomorrow?) with Ueli, if the approach is desired  resp. convenient for delivering our rpm's in context of the patch service.

chhex:
@apgsga-jhe @apgsga-uge  I will merge this branch into master tomorrow. as discussed with Ueli. Ratio: even we do not use the Yum Repo capabilities, which will be discussed with Lip Thuesday in a week, we can use it as single point of reference for the rpms. I will mail quickly when done.

PR #20, opened 2018-06-22
Java8mig353 support clone process

apgsga-jhe:
The "oc" (onClone) command of the apscli will:
- Reset the revision in Revisions.json for the corresponding target.
- Delete the corresponding artifacts (based on Revision range) from Artifactory.

I tested the following scenario, it seems ok for me (CHEI211 was configured a the productive target):

Before starting the onClone for CHEI212, Revisions.json was:

{
    "currentRevision": {
        "P": 4,
        "T": 20000
    },
    "lastRevisions": {
        "CHEI211": "3",
        "CHEI212": "10004"
    }
}

Looking at the productive BOM for revision 3 (dm-bom-9.0.6.ADMIN-UIMIG-P-3.pom), following Artifacts were "non SNAPSHOT" versions:

 <dependency>
        <groupId>com.affichage.it21.vk</groupId>
        <artifactId>zentraldispo-ui</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-2</version>
 </dependency>	  
  <dependency>
        <groupId>com.affichage.datetime</groupId>
        <artifactId>datetime-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-1</version>
 </dependency>
 <dependency>
        <groupId>com.affichage.it21</groupId>
        <artifactId>vk-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-3</version>
  </dependency>	  

After having ran onClone for CHEI212 (apscli.sh -oc CHEI212), Revisions.json was:

{
    "currentRevision": {
        "P": 4,
        "T": 20000
    },
    "lastRevisions": {
        "CHEI211": "3",
        "CHEI212": "10000@P"
    }
}

On Artifactory, "T" releases where removed from 10000 to 10004. Of course "P" releases were still there.

Then a new patch has been created for datetime-utils. When its status was "Informatiktest", content of Revisions.json was:

{
    "currentRevision": {
        "P": 4,
        "T": 20000
    },
    "lastRevisions": {
        "CHEI211": "3",
        "CHEI212": "10000"
    }
}

And the 3 above dependencies within T revision 10000 of the bom (dm-bom-9.0.6.ADMIN-UIMIG-T-10000.pom) were:

  <dependency>
        <groupId>com.affichage.it21.vk</groupId>
        <artifactId>zentraldispo-ui</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-2</version>
  </dependency>
  <dependency>
        <groupId>com.affichage.datetime</groupId>
        <artifactId>datetime-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-T-10000</version>
  </dependency>
  <dependency>
        <groupId>com.affichage.it21</groupId>
        <artifactId>vk-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-3</version>
  </dependency>
	  
Patch went then into production, Revisions.json and 3 dependencies were as follow:

{
    "currentRevision": {
        "P": 5,
        "T": 20000
    },
    "lastRevisions": {
        "CHEI211": "4",
        "CHEI212": "10000"
    }
}

  <dependency>
        <groupId>com.affichage.it21.vk</groupId>
        <artifactId>zentraldispo-ui</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-2</version>
  </dependency>
  <dependency>
        <groupId>com.affichage.datetime</groupId>
        <artifactId>datetime-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-4</version>
  </dependency>
  <dependency>
        <groupId>com.affichage.it21</groupId>
        <artifactId>vk-utils</artifactId>
        <version>9.0.6.ADMIN-UIMIG-P-3</version>
  </dependency>


apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -25,12 +29,14 @@ class PatchCli {
 	}
 		
 	public static PatchCli create() {
+		System.setProperty("logback.configurationFile", new ClassPathResource('logback.xml').path);
 		ClassPathResource res = new ClassPathResource('apscli.properties')
 		assert res.exists() : "apscli.properties doesn't exist or is not accessible!"
 		ConfigObject conf = new ConfigSlurper(profile).parse(res.URL);
 		revisionFilePath = conf.revision.file.path
```
chhex: Just thinking: Use the Config uniformly everywhere, instead of copying from the config to local def variable. This makes passing around config properties easier and also when we want to add a new one -> ala patchConfig in the Pipelines

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -621,6 +596,8 @@ class PatchCli {
 
 		def targetInd = options.rrs[0]
 		def installationTarget = options.rrs[1]
+		
+		LOGGER.info("(retrieveRevisions) Retrieving revision for target ${installationTarget} (${targetInd})")
```
chhex: I don't see why all the below logic goes into PatchCli. PatchCli should be only about: 
a. Bootstrapping
b. Parsing Commandline
c. Validating Options
d. Dispatching to respective implementing Class 

I see "important" Logic in PatchOnClone an also here.  
Probably i would have to implementing Classes: 
a. Patch Artfifactory Manager with the cleaning code.... 
b. Patch Revision Manager with the revision Logic centralized
And PatchCli dispatching to those depending on the cmd options

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -630,24 +607,59 @@ class PatchCli {
...
+		def prodTarget 
+		jsonSystemTargets.targetSystems.each{ target ->
+			if(target.typeInd == "P") {
+				prodTarget = target.target 
+			}
+		}
+		def revisions = new JsonSlurper().parseText(new File(revisionFilePath).text)
+		def prodRev = revisions.lastRevisions[prodTarget]
+		
+		// JHE (15.06.2018) : we print the json on stdout so that the pipeline can get and parse it. Unfortunately there is currently no supported alternative: https://issues.jenkins-ci.org/browse/JENKINS-26133
```
chhex: Why unfortunate?

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
```diff
@@ -2,6 +2,9 @@ package com.apgsga.patch.service.client
```
chhex: BTW: I would rename and save the revision files. If something fails, we have to analyze anyway what happened , since this operations are not "transactional" in any way.  If something is wrong we will update the file as/if required manually.  Otherwise the historical files are again something with needs to be clean up.

apgsga-jhe:
@chhex : thanks for the review. Yep agreed, I'll adjust the code and re-submit. For the logs, well, partially because one of your comment ;) ;) .... but mostly because I didn't think to log everything on "println". But yep makes sense, will also be easier to test from Eclipse (not need to have specific path configured). Yes, 100% agreed with the separation of concerns, I'll update the code accordingly. Regarding the profile: if I remember well I didn't really find any better way as what has been done so far, so yes, I like the idea of starting apscli with -D option.

chhex:
Ups , sorry  about the logging;-/ ;-) ... advantage of the groovy world , we can keep thing simpler ( mostly;-)

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
apgsga-jhe: @chhex : Actually this was done only in the context of the "-rtr" option, when we want to delete ALL Test Artifact from Artifactory. Thinking further, I believe we can/should remove this option (-rtr) from apscli, because:

1. We can easily do this from Artifactory GUI
2. In the future, we won't be able to simply remove Revisions.json file because of the production revision infos. 
3. In the future "T" releases will be implicitly cleaned up via the onClone process.

Having said that, it's really an option that we have "for us", just for now, and we can easily do this with what we already have in our hands.

What do you think?

chhex: Exactly for the moment i would keep, at least until Produktionaufnahme ;-)

apgsga-jhe: Ok, then if ok for you, I would keep the "renaming" mechanism for now. If we "forget" to remove the "-rtr" option before going live, and that someone run the "-rtr" command in production, we would loose the Revisions.json and the production revision informations.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -17,6 +19,8 @@ import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 
 class PatchCli {
+	
+	protected final static Log LOGGER = LogFactory.getLog(PatchCli.class);
 
 	// TODO JHE (12.06.2018) : can we change this with a setter? Not sure as we bootstrap the profile within create() ... if we assign a profile afterwards, obviously it will be too late.
 	public static PatchCli create(String p_profile) {
```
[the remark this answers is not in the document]
apgsga-jhe: Yep, 100% agreed that we should pass the profile differently. To make it easier (not sure it's actually easier ;) ) ... why not an apscli argument?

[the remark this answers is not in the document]
apgsga-jhe: Mmhh ok ... actually not. I forgot that we package the apscli.sh with .... probably the java -D option is better.

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCloneClient.groovy
chhex: I would make a ticket to get rid of it. Ok you could add a confirmation step.

apgsga-jhe: Good point, I created https://jira.apgsga.ch/browse/JAVA8MIG-389 .
Ok, then I'll replace the renaming by a removing :).

apgsga-jhe:
@chhex : I committed couple of modifications which should go in the direction of your review. I'm about to install this version (from this branch) on lxtwi055 to test, but I believe you can already review the last adaptations, as soon as you have some time. Regarding the separation of concerns, hopefully this is now a bit better.

apgsga-jhe:
@chhex : crap, I missed you comment yesterday (didn't receive notification :( ) ... I'm having a look right now.

PR #22, opened 2018-06-28
New location for jdbc.groovy and TargetSystemMappings.json

apgsga-jhe:
Quick review before merging .... did I forgot anything ? (or misunderstood something)


apg-patch-service-cmdclient/build.gradle
```diff
@@ -121,7 +121,7 @@ ospackage {
 	}
 	
 	from("packaging/conf") {
-		into "/var/opt/apg-patch-common"
+		into "/etc/opt/apg-patch-cli"
 		fileType CONFIG | NOREPLACE
 		include 'jdbc.groovy'
```
chhex: I think we discussed with Lip , that this would also be call ops = ops.groovy

apgsga-jhe: Oh correct, thanks. I'll change it and then merge/test.

PR #24, opened 2018-08-15
findyByIds added

apgsga-jhe:
This is what's required on service side regarding JAVA8MIG-379. Basically only a new "findByIds" method.


apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/PatchService.java
```diff
@@ -73,5 +73,9 @@
 	 */
 	public void remove(Patch patch);
 	
+	
+	public List<Patch> findByIds(List<String> patchIds);
```
chhex: Just a very very minor comment. Ordering of the method: after the findById?

apgsga-jhe: Good point, it's changed

PR #25, opened 2018-08-15
Che javamig 405 parallel builds

chhex:
Dependency Level Resolution for the Java Modules of a Patch to be used in the Pipeline for executing Builds in Parallel, when possible

The code and the approach will also be reviewed in a desk check in plenum. This pull request can serve as a preparation.

I will only merge after the review in plenum. 

Note: beside the new functionality, i also did some code clean -up and refactoring


apgsga-stb:
Hmmm... ich glaube, ich komme da nicht weit mit einem Review... oder ist die Idee, dass wir das am 16. zusammen machen?

PR #26, opened 2018-08-16
hasConflict define in Beans

apgsga-jhe:
hasconflict added to Beans


apgsga-jhe:
@chhex : on piper side ... was it also what you thought with hasConflict at Beans level?

apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/MavenArtifactBean.java
```diff
@@ -11,6 +11,7 @@
 	private String groupId;
 	private String name;
 	private String version;
+	private boolean hasConflict = false;
```
chhex: Here i would use the keyword transient

chhex: So it's transient private boolean

apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/DbObjectBean.java
```diff
@@ -11,6 +11,7 @@
 	private String fileName;
 	private String filePath;
 	private String moduleName;
+	private boolean hasConflict = false;
```
chhex: See comment below

apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/MavenArtifactBean.java
chhex: And possibly you need also the @JsonIgnore for the property

PR #28, opened 2018-08-20
"forceJadasInstallation" persisted

apgsga-jhe:
Quick branch to check the "piper side" for JAVA8MIG-381. On this side, it shouldn't be anything else that persisting the "forceJadasInstallation" property.
The other things will (probably) be managed on GUI side.


apgsga-jhe:
Yep, agreed. I renamed it, I'll merge.

PR #29, opened 2018-08-23
Manage "installJadas" field within patchBean

apgsga-jhe:
A quick one related with https://github.com/apgsga-it/jenkins-patch-scripts/pull/12

Do you see things as well like this, or am I missing something?

### DIFF
--- a/apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
+++ b/apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
@@ -152,7 +152,7 @@ public class PatchBean extends AbstractTransientEntity implements Patch {
 		final Object oldValue = Lists.newArrayList(this.dbObjects);
 		this.mavenArtifacts = mavenArtifacts;
 		firePropertyChangeEvent(MAVEN_ARTEFACTS, oldValue, mavenArtifacts);
-		setInstallJadas();
+//		setInstallJadas();
 		
 	}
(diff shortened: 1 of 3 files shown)


apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
```diff
@@ -259,7 +264,20 @@ public void setInstallOnEmptyModules(boolean installOnEmptymodules) {
 		final Object oldValue = this.installOnEmptyModules;
 		this.installOnEmptyModules = installOnEmptymodules;
 		firePropertyChangeEvent(INSTALL_ON_EMPTY_MODULES, oldValue, installOnEmptyModules);
-	}	
+		setInstallJadas();
+	}
+	
+	@Override
```
chhex: Not quite sure if this works, maybe worth a try: 
The logic (if) in the getter and a empty setter.  Should result in the groovy PatchConfig in a field with the according value

chhex: I would'nt even call setInstallJadas  (... because it does nothing) 
But as said not quite sure , if it works ;-) ... it should though, maybe you have to put the "JsonProperty" annotation on the setter and getter.

apgsga-jhe:
Ok, it worked with an empty setInstallJadas, and with the logic embedded within getInstallJadas ... all good, i'll merge.

PR #30, opened 2018-08-27
Calling OPS Script to delete Jadas Image

apgsga-jhe:
LIP provided a new script to be called in order to delete Jadas Images which won't be used anymore. 
As discussed, we will call this script at the time we clean our Releases on Artifactory.
What do you think? Do you agree with the place where the script would be called?

(we basically need a "from" and "to" parameters...)

### DIFF
--- a/apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchArtifactoryClient.groovy
+++ b/apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchArtifactoryClient.groovy
@@ -66,9 +66,11 @@ class PatchArtifactoryClient {
 			}
 			
 			// Cleaning Docker Image as well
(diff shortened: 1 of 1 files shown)


apgsga-jhe:
@chhex 
As once briefly discussed. Yep, the target doesn't actually mean anything in terms of "Jadas Image". Regarding why not calling the script from the pipeline -> it was because the pipeline doesn't know the "from" and "to" information, this is known by the apscli, who managed the Revisions.json.
I know we also spoke about eventually moving these 2 information within the json files (I believe). But it actually wouldn't help at the time of cloning, because the cleaning occurs before eventually re-installing some patches (and patch re-installation won't occur for each target).
Do you agree that we keep the call as in this Pull Request?

chhex:
@apgsga-jhe  that's fine with me

apgsga-jhe:
@chhex : dryRun variable was already there ... I just added on test ... I'll merge ...

PR #31, opened 2018-08-28
Che java8 mig 399 task executor instead of executor service

chhex:
Execute Threads with Spring TaskExecutor insteads of directorly using ExecutorServicem, see  
https://jira.apgsga.ch/browse/JAVA8MIG-399

In terms of configuration ot the TaskExecutor, question is: how much do we want to configure, see MicroServicePatchConfig.. At the moment i made some attributes configurable, but did'nt change any application.properties  = working with the defaults.

Additional some minor changes cleaning up the Code. I made a seperate Issue https://jira.apgsga.ch/browse/JAVA8MIG-417 for cleaning up JenkinsClientImpl


apgsga-jhe:
How many Threads do we want? Very good question .... Maybe not more than the number of Worker configured on Jenkins? Although it probably won't cause any issue if it's more.

chhex:
@apgsga-jhe  Concerning maxThread Configuration , i think the default 20 are (way) more then enough.  Basically the number of Threads possible must at least the number of Users currently starting a Thread at any time.  The functionality processed with the TaskExecuter resp. in Threads is: The State Transition to EntwicklungInstallationsbereit  and the Creating of new Patch Pipeline.  So i think if we have any currency ( > 1 Thread) it will be a lot. I would be astonished if we have 5 Users triggering the above functionality concurrently, but ok 2 - 3 may well be possible. That said, even though 20 is (way) to much, it doesn't hurt ;-)

PR #34, opened 2018-09-10
Stub in order to replace call to patchDbCli in Unit Test

apgsga-jhe:
Using a Stub (from Spock Framework) in order to simulate a call to a function.


apgsga-jhe:
@chhex : ok ... as briefly and indirectly discussed today ... I believe I can safely merge this one :)

PR #35, opened 2018-09-10
Searching for target.name instead of target.typeInd

apgsga-jhe:
Very minor ... but just to be sure I'm not missing anything obvious? 
I don't believe we need to change anything at any other place. Basically, the only involved function should be the one where we look for the Prod Revision.


apgsga-jhe:
@chhex : cool thanks, I'll give a try ;)

PR #52, opened 2018-10-18
Initial Iteration Apscli properties management refactoring

chhex:
Refactored the way, we config die Cli Apps.
For a first Iteration the Configuration is done by AnnotationConfigApplicationContext with the PliConfig.class.  

There you see PropertySource Configrations. which are set at Runtime as System Properties

`@PropertySource({ "${appPropertiesFile}", "${opsPropertiesFile}" })`

Basically reflecting the app and the ops properties files. 

This way testing and production configuration can be sperated without profile, see for example, the following Test Setup

```
	def setup() {
		.....
		System.properties['appPropertiesFile'] = 'classpath:config/app-test.properties'
		System.properties['opsPropertiesFile'] = 'classpath:config/ops-test.properties'
	}
```

This way the propery files have been consolidated in the two which are used for testing and the two which are deployed with the rpm. 

in the *cli.sh script instead of the 

`-Dapscli.env=production`

two 

`-DappPropertiesFile=pathToAppPropertyFile -DopsPropertiesFile=pathToOpsPropertyFile `

can be introduced.

I have'nt adopted the rpm yet.  That will be the next task. 

All test are running as before. No new tests. 
The configuration in PliConfig.class has been done in a "completed" backward compatible way. So it actually produces a Groovy ConfigObject Instance. 

I did this that, so that not to much is moved at once.  In next we can use acutally more Spring Injection with @Components. 

But not before the first produtiv Release ;-) 

I did this one,. as precondition for https://jira.apgsga.ch/browse/JAVA8MIG-469

I will not merge until reviewed and ok.


chhex:
BTW: Diff with ignoring Whitespace. 
I intially commited to master and then reset this last commit and created a Branch

chhex:
@apgsga-jhe  I revised the Spring Bootstrapping of the properties, after local rpm installation tests: 
Effectively there are now two profiles : a default for all scripts, except the DbCli scenario and one for this one. This since the right are different for app.properties and ops.properties. ops.properties are only used in the DbCli scenario.

Please have a look.

chhex:
@apgsga-jhe  see Ticket:  https://jira.apgsga.ch/browse/JAVA8MIG-477
Ich have reorganised application.properties and ops.properties in /etc/opt/apg-patch-service-server/ so that ops.properties is now empty and moved it's properties to application.properties  as result of the discussion with @apgsga-uge 
Also i renamed  app.properties  of /etc/opt/apg-patch-cli to application.properties

PR #55, opened 2018-12-04
Java8 mig 483 targetinit

apgsga-jhe:
Aim of this PR is JAVA8MIG-545. What has been done so far for JAVA8MIG-483 is not 100% ready, but I believe it doesn't hurt if we take it with ...
Pre-requisites for Unit Test to run: having an environment variable called "REPO_RO_PASSWD" set with the password for "dev" user.
Probably we'll still want to improve /apg-patch-service-cmdclient/src/main/java/com/apgsga/patch/service/client/PliConfig.java ...... for now I went ahead with the "backward compatibility solution"


apgsga-jhe:
@chhex : quick question regarding "RepositorySystemFactory". We'll make a bean out of it, no problem. Do you keep the "static" method or do we want to eliminate all statics? (sorry, don't remember) 
If possible I would eliminate the static method, right?

chhex:
@apgsga-jhe  Yep , the static methods should go away

apgsga-jhe:
@chhex : FYI, I'll close this PR without doing any merge, and will open a new one from a specific branch for JAVA8MIG-545 (as discussed). Then, when JAVA8MIG-483 will really be ready, I'll create a new PR ...

apgsga-jhe:
As written, a new one will be created for JAVA8MIG-545.

PR #56, opened 2018-12-05
Java8 mig 545

apgsga-jhe:
Pull request specific to JAVA8MIG-545. With this one we shouldn't be distracted by the "apg-patch-service-target-configinit" project (JAVA8MIG-483). Summary:

- RepositorySystemFactory is now a bean, instanciated within MicroServicePatchConfig.
- RepositorySystemFactory doesn't have "static" methods/properties anymore.
- For the bundle, it shouldn't be a problem as we'll be able to do:
`	
def repositorysystem = new RepositorySystemFactory()
repositorysystem.setRepoUser('dev')
repositorysystem.setHttpPublicArtifactoryMavenRepo('https://blabla/repo')
def artifactVersionManager = ArtifactVersionManager.create(repositories.mavenLocal().url, dmBomGroup, dmBomArtifact,patchFilePath,repositorysystem)
`

Could you please do a quick review before I merge on master?


chhex:
@apgsga-jhe  Ok, looking at it some more (locally) i am fine with the Factory
But it needs to be moved to the API : com.apgsga.artifact.query
Two more things: 
* Split the url in : Baseurl and repo
* Support password  - Here we will that take the crypt / uncrypt road for hidding it.
The problem : Unix service don't let you control enviroment variables well.

chhex:
Also in terms of naming of properties, i would prefer: 
mavenrepo.user.name
mavenrepo.user.pwd
mavenrepo.baseurl
mavenrepo.name

apgsga-jhe:
Thanks @chhex for the feedback. Part of suggestions already comitted, following is still left (but on its way ;) ):

- Also provide a default contructor and a constructor with the needed attribuates
- Support password

apgsga-jhe:
@chhex : Quick question/confirmation regarding the password. Am I right, what we want is to have the password encrypted in properties file, and decrypt it when required in our code ... right? Didn't we have this already? Do you remember for which project?

apgsga-jhe:
@chhex : I meant, what I would do is:

1. Generate a secret key.
2. Encrypt the password using this secret key.
3. Put the encrypted password in properties files.
4. Store the key in a separated file, and restrict access to this files.
5. At runtime, when needed, go and read the secret key from the file, and decrypt the password accordingly.

Do you see it also like this?

apgsga-jhe:
@chhex 
We would actually have (only difference is to read the encrypted key from a file). Also, the main would have to be ran once to generate the key and then store it into a "restricted read only" file. Or we could also integrate this step in the build process? That means the key would change from time to time ...

`public class Crypto {

	static String pwd = "hereIsThePassword";
	
	public static void main(String[] args) throws Exception {
		
		KeyGenerator keygen = KeyGenerator.getInstance("AES");
		keygen.init(256);
		SecretKey sk = keygen.generateKey();
		
		String encodedKey = Base64.getEncoder().encodeToString(sk.getEncoded());
		
		System.out.println("Encoded key : " + encodedKey);
		
		String encryptedPwd = encrypt(pwd, sk);
		System.out.println("Encrytped pwd : " + encryptedPwd);
		
		String decryptedPwd = decrypt(encryptedPwd, getSecretKey(encodedKey));
		System.out.println("Decrypted pwd : " + decryptedPwd);
		
	}
	
	private static SecretKey getSecretKey(String encodedKey) {
		byte[] decodedKey = Base64.getDecoder().decode(encodedKey);
		return new SecretKeySpec(decodedKey, 0, decodedKey.length, "AES");
	}

	private static String decrypt(String encryptedText, SecretKey secretKey) throws Exception {
		Cipher cipher = Cipher.getInstance("AES");
        Base64.Decoder decoder = Base64.getDecoder();
        byte[] encryptedTextByte = decoder.decode(encryptedText);
        cipher.init(Cipher.DECRYPT_MODE, secretKey);
        byte[] decryptedByte = cipher.doFinal(encryptedTextByte);
        String decryptedText = new String(decryptedByte);
        return decryptedText;
    }
	
	private static String encrypt(String plainText, SecretKey secretKey) throws Exception {
		Cipher cipher = Cipher.getInstance("AES");
		byte[] plainTextByte = plainText.getBytes();
		cipher.init(Cipher.ENCRYPT_MODE, secretKey);
		byte[] encryptedByte = cipher.doFinal(plainTextByte);
		Base64.Encoder encoder = Base64.getEncoder();
		String encryptedText = encoder.encodeToString(encryptedByte);
		return encryptedText;
	}
	
	

}
`

chhex:
@apgsga-jhe noooooooo ;-) haha , way to much code! We are using already using in different scenarios:  

`		<dependency>
			<groupId>org.jasypt</groupId>
			<artifactId>jasypt</artifactId>
		</dependency>`

See also http://www.jasypt.org 
And we are using the simplest Scenario the BasicTextEncryptor 
See also ConnectionBasedAuthenticationProvider in com.affichage.persistence 

`

BasicTextEncryptor textEncryptor = new BasicTextEncryptor();
textEncryptor.setPassword("realuserauth");
....
textEncryptor.encrypt((String) token.getCredentials())
`

and in StandingUserPasswCredConnectionStrategie in in com.affichage.persistence

`

BasicTextEncryptor textEncryptor = new BasicTextEncryptor();
textEncryptor.setPassword("realuserauth");
....
textEncryptor.decrypt((String) authentication.getCredentials())
`
As encryption utility : http://www.jasypt.org/cli.html#

chhex:
So it's : 

# Encrypt the password using BasicTextEncryptor with a "known" password 
# Put the encrypted password in properties files
# At runtime decypt the password form the properties files with the "hardcoded" password. 

I think that is totally enough security, what do think and if not enough, can easily be made stronger

BTW : I think Spring will be supporting encrypted property values

apgsga-jhe:
@chhex : Oh ok, effectively, much easier ;)

But still one question: in which properties files would we put the encrypted password? In an OPS file located on the server only? (Or a DEV Property file, but only present on the server)

I believe this is secure enough, no problem. But if we embedd the encrypted password into an application.properties, and store this file as part of the project on Github .... then yeah ... we actually provide everything to "crack" the password :).

Actually, thinking about it, that's also the reason why I would keep the tests running with the REPO_RO_PASSWD env variable (I believe you mentioned it once already) -> so we don't need to store pwd (encrypted or not) in order for the tests to run.

chhex:
@apgsga-jhe   "But if we embedd the encrypted password into an application.properties ..... " absolutely, yes. I was thinking about a central plattform specific properties file , instead of a ops.properties per application.  At least for the password sensitive properties. The developer szenario which could address with a file within the eclipse dist

apgsga-jhe:
@chhex 
I need your opinion regarding the "encrypted key" (not the password, but the key used to decrypt the password). At best, I believe this info will be stored in the common OPS properties file. We'll need this information within the RepositorySystemFactoryImpl class. My question is: should the RepositorySystemFactoryImpl class receive the "encrypted key" directly, or rather the path to the file where the "encrytped key" will be stored? 
At best I would pass the "encrytped key" just like the other "mavenrepo" parameter, and gives the responsibility to the caller to read the file and extract the value from there. My argument is that otherwise we need to manage a "FileNotFound" catch block within RepositorySystemFactoryImpl .... and yeah ... what do we do if the file is not found -> then we have to manage the exception in the whole chain back. But if we provide it ... we can eventually break the chain much earlier.
What do you think? Basically we would have for RepositorySystemFactoryImpl, the following constructor:

`
public RepositorySystemFactoryImpl(String mavenRepoUsername, String mavenRepoBaseUrl, String mavenRepoName, String mavenRepoUserEncryptedPwd, String keyToDecrypt) {
		this.mavenRepoBaseUrl = mavenRepoBaseUrl;
		this.mavenRepoName = mavenRepoName;
		this.mavenRepoUsername = mavenRepoUsername;
		this.mavenRepoUserEncryptedPwd = mavenRepoUserEncryptedPwd;
		this.keyToDecrypt = keyToDecrypt;
	}
`

apgsga-jhe:
@chhex 
I need your opinion regarding the "encrypted key" (not the password, but the key used to decrypt the password). At best, I believe this info will be stored in the common OPS properties file. We'll need this information within the RepositorySystemFactoryImpl class. My question is: should the RepositorySystemFactoryImpl class receive the "encrypted key" directly, or rather the path to the file where the "encrytped key" will be stored? 
At best I would pass the "encrytped key" just like the other "mavenrepo" parameter, and gives the responsibility to the caller to read the file and extract the value from there. My argument is that otherwise we need to manage a "FileNotFound" catch block within RepositorySystemFactoryImpl .... and yeah ... what do we do if the file is not found -> then we have to manage the exception in the whole chain back. But if we provide it ... we can eventually break the chain much earlier.
What do you think? Basically we would have for RepositorySystemFactoryImpl, the following constructor:

`
public RepositorySystemFactoryImpl(String mavenRepoUsername, String mavenRepoBaseUrl, String mavenRepoName, String mavenRepoUserEncryptedPwd, String keyToDecrypt) {
		this.mavenRepoBaseUrl = mavenRepoBaseUrl;
		this.mavenRepoName = mavenRepoName;
		this.mavenRepoUsername = mavenRepoUsername;
		this.mavenRepoUserEncryptedPwd = mavenRepoUserEncryptedPwd;
		this.keyToDecrypt = keyToDecrypt;
	}
`

apgsga-jhe:
@chhex : But I'm asking because it somehow looks strange to me to pass both the encrypted password and the key to decrypt .... Actually .... shouldn't we pass the decrypted password directly :) ?

chhex:
@apgsga-jhe  Thinking of the Gradle scenario. 
I would provide a RepositoryFactory.create(....). whch provides all parameters, which are needed to instanciate a Implemenation of the RepositoryFactory -> so additionally as parameters the encrypted password and and the password key. 
The caller will know where these are or how to get those. 
Gradle scenario -> via gradle-properties
Spring Boot application scenario - via ops.properties installed on target or System Enviroment properties   in development scenario. If the later causes problems, ignore it for the moment ;-))) and put them into a test.proeprties file, i don't really think it's a big deal  ..we will look at with the Eclipse packaging.

apgsga-jhe:
@chhex : yep makes sense. No problem to add an additional create method with all arguments. But I'm actually wondering what the best would be:

- A create method with all arguments.
- Or additional code within the build.gradle in order to decrypt the pwd, and use the "standard" create method.

But for now I'm going ahead with an additional create method ... It's anyway very easy to change it if we then prefer the variant with some more code in build.gradle.

chhex:
@apgsga-jhe  decrypting in responsibility of the caller: each caller needs an additional dependency and above all needs additional code for decrypting, which is same for each caller. Specially in the gradle scenario quite a burden. This way 2 x less code ;-)

apgsga-jhe:
@chhex : hehe, you probably knew it already ... but this is cool with spring. With the following, I can say "take value from property, or by default from an environment variable:

`
@Value("${mavenrepo.user.decryptpwd.key:#{environment.REPO_USER_DECRYPT_KEY}}")
private String mavenRepoUserDecryptKey;
`

I knew the default value, but didn't know until now that one could take a value from environement variable. This basically solves our Problem of storing undesired values in GIT only for Test and/or Gradle Build purpose.

apgsga-jhe:
@chhex : I just checked-in couple of adaptation, and I believe we now have a state ready to be merged on master, which will be for both JAVA8MIG-545 and JAVA8MIG-469.
Could you have a last look before I merge?

PR #59, opened 2019-03-13
JAVAMIG-703 onDemand Pipeline mit Entwickler Branch

chhex:
Siehe JAVAMIG-703

### DIFF
--- a/apg-patch-service-cmdclient/build.gradle
+++ b/apg-patch-service-cmdclient/build.gradle
@@ -20,7 +20,7 @@ targetCompatibility = 1.8
 def finalName = 'apg-patch-cli'
 
 mainClassName = 'pliStarter'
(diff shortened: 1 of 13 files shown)


apgsga-jhe:
Do we really need to have "apg-patch-service-server/db/TargetSystemMappings.json" checked-in? It doesn't hurt, but I can't find where it will be used. Maybe from the Mock Server? But then shouldn't we use the one already checked in src/test/resources/json ?

Anyway, no big deal, and looking good to me.

chhex:
@apgsga-jhe  concerning TargetSystemMapping.json: does it need to checked in? Question is actually should the directory db under source code management.  Fair enough. if it's not there - one way or other -  if you run the server locally - be it the mock or a real server configuration -  the developer needs to configure himself a runtime environment. Quite a burden.  For me it's seems quite convenient to be able the run the server out of Eclipse

PR #63, opened 2019-04-03
Avoid multiple prod patch pipeline to run in parallel when restarting a pipeline

apgsga-jhe:
Changes for java8mig-725.
I still need to test on jenkins-t, but could you already have a look, and eventually provide me a feedback?
One remark concerning com.apgsga.microservice.patch.server.impl.jenkins.JenkinsClientImpl.isLastProdPipelineBuildInError(String) method: BuildResult contains couple of other status apart from FAILURE. But in that case, we're (I believe) only interested by the FAILURE state. I mean, I don't believe we should use the "redo" if a patch was for example explicitely aborted ... in that case, standard patch workflow should be done again.


apgsga-jhe:
@chhex  , @apgsga-uge  : as said, I still need to test ... but early feedback are welcome :)

apgsga-jhe:
Good review from UGE .... isProdPatchPipelineRunning function need to deal with a Job in "Approve" state ... seems not to be considered as "isBuilding"

PR #64, opened 2019-04-04
JAVA8MIG-680 - Not waiting on feedback when submitting onDemand Job.

apgsga-jhe:
I still need to test on Jenkins-t, but as for the other Pull Request, early feedback are welcome.

Basically nothing complicated, instead of looping and waiting on the job to start, simply submit and return. Probably the "triggerPipelineJobWithoutWaitingOnFeedback" doesn't necessarily needs to be synchronized, but it also won't hurt.


apgsga-jhe:
@chhex @apgsga-uge : richtig getestet.
@apgsga-uge : man kann diese auch nächste Woche installieren, zusammen mit https://github.com/apgsga-it/piper/pull/63

PR #65, opened 2019-04-08
java8mig-730 / Performance Optimising of Multi-Patch Installation

apgsga-jhe:
Piper side for JAVA8MIG-730. Piper should provide the possibility to merge multiple patches into one big patch. When merging, patchTag should be put at MavenArtifact and DBObjects level. Once patches have been merged, the name of the newly created JSON file will be returned. The name will be prefixed with "aggregate", and will contain a datetime information in order to eventually track what happened. A name could be: Patchaggregate20190804082558.json
Not 100% sure all has been done here for piper, but probably enough in order to get at least some feedbacks ;)


apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -2,6 +2,7 @@ package com.apgsga.patch.service.client
 
 
 import org.codehaus.groovy.runtime.StackTraceUtils
+import org.codehaus.plexus.util.cli.shell.CmdShell
```
chhex: Not sure, why this dependency is here.

apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
```diff
@@ -355,4 +356,140 @@ public class PatchCliIntegrationTest extends Specification {
...
+		then:
+			result != null
+			result.returnCode == 0
+			def aggregatedPatchName = result.results.ap.aggregatePatchName
+			def Patch patch = repo.findById(aggregatedPatchName)
+			Assert.that(patch != null)
+			patch.getMavenArtifacts().size() == 11
+			patch.getMavenArtifactsToBuild().size() == 9
+			patch.getDbObjects().size() == 3
+
```
chhex: Detail: Why, not order the by artifactId and assert by position without having to loop with a case?

apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
```diff
@@ -355,4 +356,140 @@ public class PatchCliIntegrationTest extends Specification {
...
+	
+	def "Patch Cli list of patches correctly aggregated"() {
+		setup:
+			def client = PatchCli.create()
+			
+			// Ensure previous test has been cleaned up
+			Assert.that(repo.findById("aggregated") == null)
+			
+			// Persist test patches into testdb repo
+			// TODO JHE: Better write it ...
```
chhex: Yep

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
apgsga-jhe: I probably forgot a "ALT+SHIFT + o"   ;) ... Will do it

apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
apgsga-jhe: Agreed, we would then not need loops.

apgsga-jhe:
> Just a quick passthrough, apart from minor comments, the most important observations, from my point of view.
> 
> 1. I would denormalize the patch tag always , that means on save, not as special case.
> 2. Even more, important , i would separate the delivery pipeline, from the build pipeline FIRST.  This is a change on my first thoughts, agreed , but on the background  of JAVA8MIG-741, that seems to me more urgent. And there would be synergies with this one.   To be discussed.

@chhex : 
Regarding 1: Yep that would be fine for new patches, but I believe we still need to be backward compatible for a while because of existing patches, isn't it? I mean, if we aggregate a patch which hasn't been saved ... we would miss the information. But yep, in that case the aggregate function could set the patchTag only if still null or empty.
Regarding 2: I'm not sure I 100% understand what you mean. When you said "build and delivery", do you speak from the "apg-patch-script" project? or still from piper?

chhex:
@apgsga-jhe  Agreed, migration is naturally an issue.  But i would run a separate simple one off migration program  and not have migration code in piper and Jenkins pipelines proper, which after a while is "dead" code. So i would get rid this functionality altogether and not make it even more complicated ;-) Punch line: you seem to like special cases and if's  in the code, i try to avoid them , haha 
Concerning 2: Basically i am taking about our current Patch Pipeline, which now build , assembles and installs on a patch basis.  The suggestions to have two separate "Patch" Pipelines , one for building on a Patch Basis. And one for Assembly and Installation , scheduled, which aggregates the Patches to be Installed. So the It21 wouldn't be installed anymore on a Patch basis. 

But maybe let's discus this later on..

apgsga-jhe:
@chhex : :) :) ..... well, not that I like "ifs" that much :) .... but I don't like to "manually" change the production json patch files .... but yeah, "ifs" which will become deprecated is not a better choice, agreed.

chhex:
@apgsga-jhe I think we can close this pull request.

PR #66, opened 2019-04-10
JAVA8MIG-742 last successful Pipeline Task field on API

chhex:
Neues Feld in Patch: lastPipelineTask, siehe auch https://github.com/apgsga-it/jenkins-patch-scripts/pull/47

### DIFF
--- a/apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
+++ b/apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
@@ -33,6 +33,7 @@ public class PatchBean extends AbstractTransientEntity implements Patch {
 	private List<DbObject> dbObjects = Lists.newArrayList();
 	private List<MavenArtifact> mavenArtifacts = Lists.newArrayList();
 	private boolean installOnEmptyModules = false;
+	private String lastPipelineTask ="";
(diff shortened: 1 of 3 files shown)


apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/impl/PatchBean.java
```diff
@@ -313,6 +314,17 @@ public String getTargetToState() {
 	public void setTargetToState(String targetToState) {
 		this.targetToState = targetToState;
 	}
+	
+	
+	@Override
+	public String getLastPipelineTask() {
+		return targetToState;
```
apgsga-jhe: Shouldn't we return "lastPipelineTask" ?

chhex: @apgsga-uge  good point

chhex: @apgsga-jhe @apgsga-uge done

chhex:
@apgsga-jhe @apgsga-uge  I mergeing these change to master.

chhex:
@apgsga-jhe @apgsga-uge with a bumped version.

PR #67, opened 2019-04-11
separated onclone pipeline for each target

apgsga-jhe:
Not sure the pull request name is 100% right ... is it really related with JAVA8MIG-730 ... but ok, detail...

- Calling onClone pipeline with target specific name
- Catching exception when fetching list of patches to be reinstalled and printing "true"/"false" on stdout in order for the caller to catch the result.
(- Ignoring one test as it still needs to be adapted for JAVA8MIG-725...)

### DIFF
--- a/apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
+++ b/apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
@@ -121,8 +121,6 @@ public class PatchCliIntegrationTest extends Specification {
 		repo.clean()
 	}
 	
-	// JHE (11.04.2019) : Deactivate for now, needs to be adapted based on changes made for JAVA8MIG-725
-	@Ignore
 	def "Patch Cli redo's Patch, which has been saved before with -sa"() {
 		setup:
 		def client = PatchCli.create()
--- a/apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
+++ b/apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
@@ -1,17 +1,13 @@
 package com.apgsga.patch.service.client;
 
-import org.junit.Rule
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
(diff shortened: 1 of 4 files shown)


apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
```diff
@@ -127,6 +121,8 @@ public class PatchCliIntegrationTest extends Specification {
 		repo.clean()
 	}
 	
+	// JHE (11.04.2019) : Deactivate for now, needs to be adapted based on changes made for JAVA8MIG-725
+	@Ignore
```
apgsga-uge: I didn't know this way of deactivation, so I looked it up and found some comments proposing not to use it.
Since it's just a temporary deactivation, I would agree to do it just with @Ignore and not to invest mor time on the topic…
just for info about alternatives: https://www.codeaffine.com/2013/11/18/a-junit-rule-to-conditionally-ignore-tests/

apgsga-uge:
for me it looks like what we need :-)

apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
apgsga-uge: ... just as additional info:
I placed this comment in order to see what the result would be in terms of visualisation within the conversation on github - so, if you don't like it, then just @ignore ;-)

chhex: @apgsga-jhe  @apgsga-uge  I don't think the test case needs  to be adopted. But  the /apg-patch-service-server implementation of JAVA8MIG-725.  I a little astonished that the change passed, because basically all regression test should run ok. What obviously is missing is a mock implementation for the change. The tests here are about parameterization and the server call of apscli, see also @ActiveProfiles on top the the Test class.

apgsga-jhe:
> As discussed in JAVA8MIG-730, this change do not have anything to do with initial intent of the ticket.

@chhex : Name of pull request changed.

apg-patch-service-cmdclient/src/test/groovy/com/apgsga/patch/service/client/PatchCliIntegrationTest.groovy
apgsga-jhe: @chhex : good point, following mock implementation within JenkinsMockClient should actually return true, and not false:

`	@Override
	public boolean isLastProdPipelineBuildInError(String patchNumber) {
		LOGGER.info("isLastProdPipelineBuildInError for : " + patchNumber);
		return true;
	}	`

apgsga-jhe:
@chhex , @apgsga-uge  : I'm merging on master ...

PR #68, opened 2019-04-16
jhe-cm-176 / distinguished patchToBeReinstall file names

apgsga-jhe:
JSON file with list of patch to be reinstalled needs to have distinguished name for each status (informatiktest, anwendertest, ...)


apgsga-jhe:
@chhex , @apgsga-uge : JFYI .... I'll merge immediately. Background is that we need the list of patch to be reinstalled to be rendered in dedicated files.

apgsga-jhe:
> DbCliIntegrationTest.groovy is going to create an output file with hard coded target status. Since there are different target status' depending on the environment to be installed, this could hinder the flexibility of the workflow

Correct, but this is a test where we want to explicitly check what the output will be. But yep, we could better write the test and fetch the status from the test TargetSystemMapping file.

apgsga-uge:
> DbCliIntegrationTest.groovy is going to create an output file with hard coded target status. Since there are different target status' depending on the environment to be installed, this could hinder the flexibility of the workflow
> 
> @apgsga-uge
> Correct, but this is a test where we want to explicitly check what the output will be. But yep, we could better write the test and fetch the status from the test TargetSystemMapping file.

oh - right you are... sorry

PR #70, opened 2019-04-25
java8mig-747 / Preconditions to allow redo of aborted Pipelines

apgsga-jhe:
In order not to have unwanted behaviors, I kept the pre-condition with the "failed" pipeline, but added as well the "aborted" scenario.
What do you guys think? 
Ideally we could patch this one together with JAVA8MIG-732, where we'll anyway have to install a new version of Piper.

### DIFF
--- a/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/jenkins/JenkinsClientImpl.java
+++ b/apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/jenkins/JenkinsClientImpl.java
@@ -315,23 +315,14 @@ public class JenkinsClientImpl implements JenkinsClient {
 	}
 
 	@Override
-	public boolean isLastProdPipelineBuildInError(String patchNumber) {
-		return matchLastProdPipelineBuildResult(patchNumber, BuildResult.FAILURE);
-	}
-
-	@Override
-	public boolean isLastProdPipelineBuildAborted(String patchNumber) {
-		return matchLastProdPipelineBuildResult(patchNumber, BuildResult.ABORTED);
-	}
-	
-	private boolean matchLastProdPipelineBuildResult(String patchNumber, BuildResult expectedBuildResult) {
+	public BuildResult getProdPipelineBuildResult(String patchNumber) {
 		JenkinsServer jenkinsServer;
 		try {
 			jenkinsServer = new JenkinsServer(new URI(jenkinsUrl), jenkinsUser, jenkinsUserAuthKey);
 			PipelineBuild lastBuild = getPipelineBuild(jenkinsServer, patchNumber);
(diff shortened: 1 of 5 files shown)


apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/impl/jenkins/JenkinsClientImpl.java
```diff
@@ -316,13 +316,22 @@ public boolean isProdPatchPipelineRunning(String patchNumber) {
...
 	public boolean isLastProdPipelineBuildInError(String patchNumber) {
+		return matchLastProdPipelineBuildResult(patchNumber, BuildResult.FAILURE);
+	}
+
+	@Override
+	public boolean isLastProdPipelineBuildAborted(String patchNumber) {
+		return matchLastProdPipelineBuildResult(patchNumber, BuildResult.ABORTED);
+	}
+	
+	private boolean matchLastProdPipelineBuildResult(String patchNumber, BuildResult expectedBuildResult) {
```
chhex: The API doesn't support conditions of Build Results. So which each expected build you need to make a round trip to jenkins. If you what to do this generic , then the api should be a little more flexible.

chhex: Or otherwise keep it simple: you return the BuildResult and check the condition where needed.

apgsga-jhe: Yep, I changed it in order to return the BuildResult.

apgsga-uge:
> In order not to have unwanted behaviors, I kept the pre-condition with the "failed" pipeline, but added as well the "aborted" scenario.
> What do you guys think?
> Ideally we could patch this one together with JAVA8MIG-732, where we'll anyway have to install a new version of Piper.

I'd suggest anyway, that we coordinate the update with UMB (sch or ber) in order to let them produce a snapshot of the linux machine (as backup) just before updating - this way we could at least "jump" back if we encounter any Problems after installation

apgsga-jhe:
> > In order not to have unwanted behaviors, I kept the pre-condition with the "failed" pipeline, but added as well the "aborted" scenario.
> > What do you guys think?
> > Ideally we could patch this one together with JAVA8MIG-732, where we'll anyway have to install a new version of Piper.
> 
> I'd suggest anyway, that we coordinate the update with UMB (sch or ber) in order to let them produce a snapshot of the linux machine (as backup) just before updating - this way we could at least "jump" back if we encounter any Problems after installation

Actually, by unwanted behaviors, I meant when executing the "redo" :) . But in any case, I find it a very good idea if we can have a SNAPSHOT before.

PR #71, opened 2019-04-26
java8mig-748 / IT21-CM-GUI: Suche nach bestimmten Java-Artifakten

apgsga-jhe:
Ability to get list of Patches containing a given module name. Module can either be a Java Artifact of a DB Module.
As in code commented, we might want to change on which field the research will be made, this can easily be change. For now, the research would be done:
- On ArtifactId for MavenArtifact objects
- On ModuleName for DBObject
GUI will obviously also have to be updated, but this one can (has to) go in production first.


apgsga-jhe:
@chhex , @apgsga-uge  : I hope you guys don't mind .... but I already merge this one on master. I'm actually ready to test from the GUI, and would like to install a new version on jenkins-t. Obviously, we can still change things directly on master before doing the final merge on prod branch.

PR #85, opened 2019-10-30
CM-243 / sources JARs added in search of Artifacts to be deleted

apgsga-jhe:
"sources" JARs not found in our requests


apgsga-jhe:
> why not removeArtifacts("*${revisionFormatedForSearch}*",...) ?

@apgsga-uge : because I'm scared to delete resources which shouldn't be :) . For example, if searching for Artifacts which have the revision number 4041, you'll find (non-exhaustive list):

1. apg-jadas-service-CHTI211-4041.6201-1.noarch.rpm
2. patch_6201-ADMIN-UIMIG-4041.zip
3. bd-dao-9.0.6.ADMIN-UIMIG-20180613.140413-22.jar

The first 2 are fine, they have to be deleted. The last one shouldn't have to be deleted .... Therefore, i'm addin a "-" and "." in my criteria. 

Ok, there is a bit of paranoia, because the one which shouldn't be deleted should be in a different repository .... But I took an approach where I prefere beeing a bit too restrivitve, rather than deleting resources which shoulnd't be.

PR #90, opened 2020-01-29
IT-34991 , IT21 cm-ui, SQL Objekte brauchen Anpassungen um selektierbar zu sein

apgsga-jhe:


build.gradle
```diff
@@ -9,7 +9,7 @@ project.ext {
 allprojects  {
 	apply plugin: 'maven'
 	group = 'com.apgsga.patchframework'
-	version  = '1.3.24'
+	version  = '1.3.44'
```
apgsga-uge: are you sure about 1.3.**4**4 ?

apgsga-jhe: > are you sure about 1.3.**4**4 ?

no no, this will be changed just before merging, but for now this is the number for jenkins-t.

apg-patch-service-client/src/main/java/com/apgsga/microservice/patch/client/MicroservicePatchClient.java
```diff
@@ -142,6 +144,16 @@ public void startInstallPipeline(Patch patch) {
 		return Lists.newArrayList(result);
 	}
 
+	@Override
+	public List<DbObject> listAllSqlObjectsForDbModule(String patchNumber, String searchString, String username) {
```
apgsga-uge: it's not exactly ...SqlObjects... any longer - but never mind

apgsga-jhe: @apgsga-uge : no no, good point. I'm actually anyway busy with some modification, let's name this correctly as well. listAllObjectsForDbModule ? what do you think?

apgsga-jhe: @apgsga-uge : now it's ready for tomorrow, just the version will have to be updated ...

apgsga-jhe:
@apgsga-uge : I changes the version to match next production installation. I'll now merge so that rpm are ready for the installation this morning, and also to go forward with the GUI (I'll have to change something there as well, and need the newest version).

PR #92, opened 2020-04-08
Multiservice cm First Iteration

chhex:
@apgsga-jhe  Can you have a quick look? I  refactored com.apgsga.microservice.patch.api.impl.PatchBean into com.apgsga.microservice.patch.api.Patch basically getting rid of the Interface. 
This should make changes easier.
Tradeoff , we need to migrate our Patch*.json Files. 
But then again we need to do that any way, i do not see anyway around.
What do you think?

Except you shout loud and soon , i will go ahead an migrate the other Beans in com.apgsga.microservice.patch.api.impl


apgsga-jhe:
@chhex : as discussed ;) ... fine for me

PR #95, opened 2020-04-08
Multiservice cm 2

chhex:


chhex:
@apgsga-jhe Can you have a quick look? I refactored com.apgsga.microservice.patch.api.impl.PatchBean into com.apgsga.microservice.patch.api.Patch basically getting rid of the Interface.
This should make changes easier.
Tradeoff , we need to migrate our Patch*.json Files.
But then again we need to do that any way, i do not see anyway around.
What do you think?

Except you shout loud and soon , i will go ahead an migrate the other Beans in com.apgsga.microservice.patch.api.impl

chhex:
@apgsga-jhe  First functional Version as Basis for the Jenkins development and the Development of the new UI. 

All test cases pass. Effectively the changes weren't really that dramatic. 

Main areas of change: 

- Version Resolving for Revision Handling removed, since this is now addressed by a Gradle Plugin 
- The Patch API changed to support Multiple Services, see also https://github.com/apgsga-it/piper/blob/Multiservice_cm_2/apg-patch-service-common/src/main/doc/patch.png
- Removed unnecessary separation of interface and implementation of the Patch API Beans. 

The changes should be backward compatible with regard to areas , which are not effected by Multiservice CM. 

We can merge this one into Master , if you want to? 

@apgsga-stb FYI

apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/DbObject.java
```diff
@@ -1,33 +1,118 @@
 package com.apgsga.microservice.patch.api;
 
-import com.affichage.persistence.common.client.EntityRootInterface;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.affichage.persistence.common.client.AbstractTransientEntity;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.apache.commons.io.FilenameUtils;
```
apgsga-jhe, line 6: @chhex : Probably minor, but what is the reason of this change? (AbstractTransientEntitiy instead of @EntityRootInterface) Maybe just that we are now using the latest persistence-framework version? I noticed the same further down for "MavenArtifact", "PatchLog" and "PatchLogDetails".

apgsga-jhe:
@chhex : Apart one question (which is more a "curiosity question"), this is looking good to me. To my point of view you can merge it on master.

apg-patch-service-common/src/main/java/com/apgsga/microservice/patch/api/DbObject.java
chhex: @apgsga-jhe  The @EntityRootInterface Interface , was need for Forms2Java. The Annotation tells the Framework : i am a Bean, please consider my setters and getters.  The Root means: I am an "Entrypoint" for property resolution. Since we are not using Interfaces an more, this has become obsolete.  AbstractTransientEntity actually provides the Property Change Listeners. So all our Piper Beans , which are relevant for the UI inherit from AbstractTransientEntity. Transient because, they don't use the standard It21 persistence mechanism.

apgsga-jhe: @chhex : (y), thanks for the explanations

chhex: @apgsga-jhe  I made some amendments to the text, should be even clearer now.

PR #96, opened 2020-04-21
Refactored Piper Modules 

chhex:
- Renamed apg-patch-service-common to apg-patch-service-api
- Moved core functionality from apg-patch-service-server to apg-patch-service-core

Testcases run ok. 

Next steps: 

- Optional dependency apg-patch-service-cmdclient to  apg-patch-service-core
- Serverless PatchServiceClient Implementation Variant, which calls apg-patch-service-core
- Configuration Option for Serverless PatchServiceClient
- Zip Packaging Variant for Serverless PatchServiceClient 
- Remove Revision Code in apg-patch-service-cmdclient


apg-patch-service-server/src/main/java/com/apgsga/microservice/patch/server/MicroPatchServer.java
```diff
@@ -3,10 +3,11 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.ComponentScan;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
 
 @SpringBootApplication
-@ComponentScan(basePackages = { "com.apgsga.microservice.patch.server" })
-
+@ComponentScan(basePackages = { "com.apgsga.microservice.patch.server", "com.apgsga.microservice.patch.core" })
+@EnableWebMvc
```
apgsga-jhe, line 10: @chhex : No problem at all for me, but out of curiosity, why do we need EnableWebMvc? I thought Springboot would enable it himself if he notice the required jars on the classpath, but not sure of that.

chhex: @apgsga-jhe Good point, not sure neither. I actually just refactored it. I moved it from the SimplePatchContainerBean @Component, since that inhibited the reused out of a Servlet container.

chhex: @apgsga-jhe I will take it away and look what happens ;-)

chhex: @apgsga-jhe  Apart from that , it is totally Work in progress. About that later with Teams.

PR #98, opened 2020-06-25
IT-35956 : Notification back to the main pipeline

apgsga-jhe:
How could we send notification back to main pipelines that a step has been done.


chhex:
@apgsga-jhe  yes, that is more or less obvious and what i also was thinking. But i would say, we are fare away from this in terms of priority ;-)

apgsga-jhe:
Closing without merging for now.

PR #99, opened 2020-06-25
Rpm & Gem Packaging for Serverless apscli with clean-up

chhex:


apgsga-jhe:
@chhex  : I believe we can close this one without merging anything, right?

chhex:
No, leave that one, because of the gem packageing which i started they .... just point of reference

chhex:
Obsolete

PR #103, opened 2020-07-14
IT- 35961 / Job Preconditions and Submit API for install pipeline

apgsga-jhe:
@chhex : One which shouldn't be complicated (hopefully ;) ).
Basically the Piper APIs in order to start our future "install" Pipelines.


apgsga-jhe:
@chhex : I'm merging this one now before having branches which will go in different ways ... It as nothing big, basically having the possibility to start an install pipeline from apscli.

PR #104, opened 2020-08-12
IT-35944 / Move Targetsystem Mapping in Pipeline Repo (Pipeline Libs, Piper)

apgsga-jhe:
Name of the branch is wrong .... it should have been jhe_it-35944 .... but anyway ....

@chhex : here a very first sketch of where the TargetSystemMapping could be, and a very first sketch of the API needed at the moment. Obviously not ready to be merged ;) ... but, what do you think?


apg-patch-system-mapping/src/main/java/com/apgsga/system/mapping/api/TargetSystemMapping.java
```diff
@@ -0,0 +1,35 @@
+package com.apgsga.system.mapping.api;
```
chhex, line 1: @apgsga-jhe 
yep, looks good. Concerning this Interface.... it will be only used in the Java scenario. So i would concrete on those use cases ....

apg-patch-system-mapping/src/main/java/com/apgsga/system/mapping/impl/TargetSystemMappingImpl.java
```diff
@@ -0,0 +1,47 @@
+package com.apgsga.system.mapping.impl;
```
chhex, line 1: @apgsga-jhe  Here in implementation , i would expect a Spring managed Component , which serves as facade to the Pipeline Library. and executes the code accordingly. 
Basically is should be Singleton which upon initialization clones the GitHub library and loads the Script.
The methods are then actually execution of functions of the script , with according parametrization , see also https://github.com/apgsga-it/piper/blob/master/apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/GroovyScriptActionExecutor.java and http://docs.groovy-lang.org/latest/html/documentation/guide-integrating.html

apg-patch-system-mapping/src/main/java/com/apgsga/system/mapping/api/TargetSystemMapping.java
chhex: @apgsga-jhe  i would for the moment do with findStatus

apgsga-jhe: > @apgsga-jhe i would for the moment do with findStatus

@chhex : not sure I understand this one :)

apg-patch-system-mapping/src/main/groovy/com/apgsga/system/mapping/targetSystemMappings.groovy
```diff
@@ -0,0 +1,52 @@
+package com.apgsga.system.mapping
```
chhex, line 1: @apgsga-jhe yep, looks good.

I would necessarily use the Jenkins function for reading the "Library" file , see https://github.com/apgsga-it/apg-jenkins-pipeline-testlibs/blob/master/src/main/jenkins/declarativeWithTypeMappingOnly.groovy

Hmmmm that said, that only works in a Jenkins context ;/
Revising what i wrote before: i would pass the content of the File  as Text to the create method and leave the reading outside of the scope of the Singleton

In terms of Groovy Singleton , did you check with examples?

apgsga-jhe: @chhex : Yep, agreed regarding using the Jenkins function for reading targetSystemMappings.groovy, I'll move the file in a "resources" folder, within the same Piper's module as it is now. If I understood it well, that should be enough for Jenkins to find i using its standard function.

Yes, we could also pass the content of the File as parameter, instead of the file path. But .... I don't see why this wouldn't work if we pass the file path?

Regarding the Singleton, I made it based on the following: https://docs.groovy-lang.org/latest/html/api/groovy/lang/Singleton.html
From what I understood, this is enough to annotate the classe with @Singleton, and to instantiate it with "<className>.instance".

apgsga-jhe:
@chhex : a first version of the Groovy TargetSystemMapping would be ready for a first review/feedback. Piper in general still have to be adapted, but the component itself could already be reviewed.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/config/MicroServicePatchConfig.java
```diff
@@ -2,6 +2,8 @@
 
 import java.io.IOException;
 
+import com.apgsga.system.mapping.api.TargetSystemMapping;
+import com.apgsga.system.mapping.impl.TargetSystemMappingImpl;
```
chhex, line 6: @apgsga-jhe Here i would try to " isolate" all operations with regard to TargetSystemMapping. 
Actually MicroservicePatchConfiug shouldn't need to know about com.apgsga.system.mapping.impl.TargetSystemMappingImpl 
or any configuration of TargetSystemMapping 
I would make TargetSystemMapping a Spring @Component in it's own right , which is autowind where needed.

apgsga-jhe:
> @apgsga-jhe Sorry , with being late....
> I was thinking: you created a separate module for TargetsystemMapping......
> I think that would be even more so a good approach of the DbModule.....
> in terms of the dependencies and seperation of concerns
> 
> What do think?

@chhex : no no, no problem, you're not late, this pull request will at the end be for more than one ticket (it's difficult to separate all refactoring into different pull request). So ... if you already see something, feedbacks are more than welcome, but everything is not finish yet.

Yep, good point. Effectively, I created an "apg-patch-system-mapping" module as part of Piper project. For now I separated the DB stuffs into separated packages, but yeah, a separated module wouldn't be bad (y). I'll move the code, that shouldn't be a big deal.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/GroovyScriptActionExecutor.java
```diff
@@ -1,73 +1,48 @@
...
-
-import groovy.lang.Binding;
-import groovy.lang.GroovyShell;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.core.io.support.ResourcePatternResolver;
 
+import java.io.IOException;
+import java.io.InputStream;
+
 public class GroovyScriptActionExecutor implements PatchActionExecutor {
```
chhex, line 19: @apgsga-jhe  You are surely aware that PatchActionExector is used also otherwise in Piper? 
I wouldn't change anything on that one , but have Code in or via com.apgsga.system.mapping.impl.TargetSystemMappingImpl which executes the needed Groovy dynamically
As said above, i would try if possible to have com.apgsga.system.mapping.impl.TargetSystemMappingImpl  a Spring Component in it's own right.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/config/MicroServicePatchConfig.java
apgsga-jhe: > @apgsga-jhe Here i would try to " isolate" all operations with regard to TargetSystemMapping.
> Actually MicroservicePatchConfiug shouldn't need to know about com.apgsga.system.mapping.impl.TargetSystemMappingImpl
> or any configuration of TargetSystemMapping
> I would make TargetSystemMapping a Spring @component in it's own right , which is autowind where needed.

@chhex : Yep. Actually, the reason why I added it here was because I need a TargetSystemMapping when instantiating the  groovyActionFactory bean (within MicroServicePatchConfig class). However, agreed, I could Autowired it within MicroServicePatchConfig (I believe) -> I'll give a try.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/GroovyScriptActionExecutor.java
apgsga-jhe: > @apgsga-jhe You are surely aware that PatchActionExector is used also otherwise in Piper?
> I wouldn't change anything on that one , but have Code in or via com.apgsga.system.mapping.impl.TargetSystemMappingImpl which executes the needed Groovy dynamically
> As said above, i would try if possible to have com.apgsga.system.mapping.impl.TargetSystemMappingImpl a Spring Component in it's own right.

@chhex : Oh crap, I was so proud to reduce code in this area ;) , ahahaha .... Mmhh, ok, no, I though it was always used in context of Piper. Then yeah of course, I'll change the approach there.

apgsga-jhe:
> > @apgsga-jhe Sorry , with being late....
> > I was thinking: you created a separate module for TargetsystemMapping......
> > I think that would be even more so a good approach of the DbModule.....
> > in terms of the dependencies and seperation of concerns
> > What do think?
> 
> @chhex : no no, no problem, you're not late, this pull request will at the end be for more than one ticket (it's difficult to separate all refactoring into different pull request). So ... if you already see something, feedbacks are more than welcome, but everything is not finish yet.
> 
> Yep, good point. Effectively, I created an "apg-patch-system-mapping" module as part of Piper project. For now I separated the DB stuffs into separated packages, but yeah, a separated module wouldn't be bad (y). I'll move the code, that shouldn't be a big deal.

@chhex : We now have a module for DB: https://github.com/apgsga-it/piper/pull/104/commits/e0c83b63f3ab4e70cd6cc3cd6248caddecb65c1d

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/GroovyScriptActionExecutor.java
apgsga-jhe: @chhex : I searched for places where PatchActionExecutor is used in Piper, and can't find a place where the refactoring would cause a problem. I'll adapt the MicroservicePatchConfig in order to to create a bean for "TargetSystemMapping", but I don't believe this should cause a problem with the refactoring done in context of the PatchActionExecutor implementation. 
Or did you mean that it is called as well outside from Piper, for example directly from our Pipeline? (I don't believe)

apgsga-jhe:
@chhex : not sure how deep you want to review this one, since it involves couple of changes ... It basically covering following tickets:

- IT-35944
- IT-36187
- IT-36189
- IT-36203

I would like to merge it before starting IT-35951.

For me it's also fine if you don't review it that deep. If we realize that things are missing when we'll integrate our Pipelines, we can then correct it ....

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/config/PliConfig.java
```diff
@@ -20,7 +20,7 @@
```
chhex, line 20: @apgsga-jhe  I would strongly consider to totally get rid of PliConfig, and provide the target host via cmdline option -  possibly  optionally, if not specified assume it is set via SystemProperty. The last scenario for the Pipeline usage, where  we can configure the Target via Environment.  That way no property file is needed anymore.  So the apscli is really totally "moveable"

apgsga-jhe: @chhex : (y), ok yep, will try to completely remove it.

apgsga-jhe: @chhex : Here we go, apscli without pliConfig: https://github.com/apgsga-it/piper/pull/104/commits/c6f5abe4567eca6a496e3f35081ab55733353a65
I'll soon merge ...

PR #105, opened 2020-08-26
ProcessBuilder Examples

chhex:
@apgsga-jhe  see ProcessRunnerTest
A couple of remarks:

- Basically the stuff seems to working as expected.
- Ok, i did't test on Windows, so the ssh invocation will be different , see static String List SSH_CMD, probably some more commands
- I am running against a JENKINS_HOST and running a couple of ssh jenkins cli commands
- The naming is not good in the current Implementation of ProcessBuilderCmdRunner et al. VcsCommand should really be something like Command or ProcessCommand
- First there are some tests just reusing the ProcessBuilderCmdRunner.StreamGobbler
- And then one Test using ProcessBuilderCmdRunner directly


apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/vcs/ProcessBuilderCmdRunner.java
```diff
@@ -13,66 +13,78 @@
...
+        }
+
+        @Override
+        public void run() {
+            try {
+                InputStreamReader isr = new InputStreamReader(is);
+                BufferedReader br = new BufferedReader(isr);
+                String line;
+                while ((line = br.readLine()) != null) {
+                    LOGGER.info(line);
```
chhex, line 56: @apgsga-jhe  Not quite sure why we are doing this logging here

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/vcs/ProcessBuilderCmdRunner.java
```diff
@@ -13,66 +13,78 @@
...
+            } catch (IOException ioe) {
+                LOGGER.error(ioe);
+            }
+        }
+
+        public List<String> getOutput() {
+            return output;
+        }
+
+        public void log() {
```
chhex, line 68: @apgsga-jhe Added a logging on demand

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/vcs/ProcessBuilderCmdRunner.java
```diff
@@ -13,66 +13,78 @@
...
+            if (exit != 0) {
+                throw new AssertionError(String.format("ProcessBuilder returned ExitCode %d", exit));
+            }
+            return Lists.newArrayList(outputGobbler.getOutput());
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static class StreamGobbler extends Thread {
```
chhex, line 40: @apgsga-jhe  Changed to public so the Gobbler can be used isolated

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+public class ProcessRunnerTest {
+
+    public static final String USER = "che";
+    public static final String JENKINS_HOST = "172.16.92.196";
+    public static final String JOB_NAME = "TestappBom";
+    public static List<String> SSH_CMD = Lists.newArrayList("ssh");
+
+    @Test
+    public void testJenkinsCliHelp() throws Exception {
```
chhex, line 18: @apgsga-jhe Simple Help , without the ProcessBuilderCmdRunner, but using the Gobbler

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        int exitCode = process.waitFor();
+        errorGobbler.join();
+        outputGobbler.join();
+        outputGobbler.log();
+        errorGobbler.log();
+        org.junit.Assert.assertEquals("No errors should be detected", 0, exitCode);
+    }
+
+    @Test
+    public void testJenkinsCliBuildJob() throws Exception {
```
chhex, line 37: @apgsga-jhe Build a Job and wait until started

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        int exitCode = process.waitFor();
+        errorGobbler.join();
+        outputGobbler.join();
+        outputGobbler.log();
+        errorGobbler.log();
+        org.junit.Assert.assertEquals("No errors should be detected", 0, exitCode);
+    }
+
+    @Test
+    public void testJenkinsCliConsole() throws Exception {
```
chhex, line 56: @apgsga-jhe Get output of the Last Build of the Job , follow if running

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        errorGobbler.join();
+        outputGobbler.join();
+        outputGobbler.log();
+        errorGobbler.log();
+        // This seems to be a problem with jenkinscli
+        org.junit.Assert.assertEquals("No errors should be detected", 255, exitCode);
+    }
+
+    @Test
+    public void testJenkinsCliStartBuildAndFollow() throws Exception {
```
chhex, line 76: @apgsga-jhe Both above : Start Build Job , wait until started and follow output immediately

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        outputGobbler.join();
+        outputGobbler.log();
+        errorGobbler.log();
+        // This seems to be a problem with jenkinscli
+        org.junit.Assert.assertEquals("No errors should be detected", 255, exitCode);
+
+    }
+
+    @Test
+    public void testJenkinsCliHelpWithProcessBuilderCmdRunner() {
```
chhex, line 112: @apgsga-jhe Help with the ProcessBuilder......which does reduce the code.

apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/impl/cmd/ProcessRunnerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        ProcessBuilderCmdRunner.StreamGobbler errorGobbler = new ProcessBuilderCmdRunner.StreamGobbler(process.getErrorStream());
+        ProcessBuilderCmdRunner.StreamGobbler outputGobbler = new ProcessBuilderCmdRunner.StreamGobbler(process.getInputStream());
+        outputGobbler.start();
+        errorGobbler.start();
+        int exitCode = process.waitFor();
+        errorGobbler.join();
+        outputGobbler.join();
+        outputGobbler.log();
+        errorGobbler.log();
+        // This seems to be a problem with jenkinscli
```
chhex, line 71: @apgsga-jhe There seems to be an issue with here with the jenkins cli over ssh. Not a big one. But the Process terminates with 255 , instead of 0, because of:

`org.apache.sshd.common.SshException: flush(ChannelOutputStream[ChannelSession[id=0, recipient=0]-ServerSessionImpl[che@/172.16.92.1:60271]] SSH_MSG_CHANNEL_DATA) length=0 - stream is already closed
	at org.apache.sshd.common.channel.ChannelOutputStream.flush(ChannelOutputStream.java:169)
	at org.jenkinsci.main.modules.sshd.AsynchronousCommand$1.run(AsynchronousCommand.java:114)
	at java.lang.Thread.run(Thread.java:748)`

chhex: See also https://issues.jenkins-ci.org/browse/JENKINS-31194

apgsga-jhe:
@chhex : Thanks for the example, this is really useful, and basically a very good basis for our meeting tomorrow. So, that means the multi-threading integration was a problem. Do you think we should try to have 1-2 tests outside of Piper (and Springboot), but with couple of threads started in parallel? 
Probably we'll be able to fulfill most of our requirements with "build" and "console", knowing that we might have to parse the output received from a "build" job (in order for example to get the build number, which will then be passed to the "console"). But Parsing the build-output shouldn't be a big problem I believe.

chhex:
@apgsga-jhe
 I don't think the multithreading was directly the problem, but the main thread waiting for the command thread's to finish. A typical integration test problem with spring boot , since the startup , shutdown sequence is done relatively quick.  So i don't think that we need additional tests here for the moment. If you want to stick with the integration test scenario , some additional thoughts are needed.  Working with count down latches and waits etc. Also the running of the command needs to be guarded against flakiness .... not easy. 

I wouldn't parse the output for the last build number: The Patch Job should be configured that only one Job runs and since we are waiting with -w for the job to start, we should be fine with the lastBuild Output.

apgsga-jhe:
@chhex : In the meantime I could try to run your test examples on my side, they're working fine. In the output window, I only see the following info for tests involving the "console" command:

```
08:48:45.581 [Test worker] INFO com.apgsga.microservice.patch.core.impl.vcs.ProcessBuilderCmdRunner - org.apache.sshd.common.SshException: flush(ChannelOutputStream[ChannelSession[id=0, recipient=0]-ServerSessionImpl[jhe@/172.17.139.241:49518]] SSH_MSG_CHANNEL_DATA) length=0 - stream is already closed
	at org.apache.sshd.common.channel.ChannelOutputStream.flush(ChannelOutputStream.java:169)
	at org.jenkinsci.main.modules.sshd.AsynchronousCommand$1.run(AsynchronousCommand.java:114)
	at java.lang.Thread.run(Thread.java:748)
```

But I don't believe this is something to be worried about.

No no, I don't have to have these integration tests. When I wrote them, I didn't think of the Springboot problem in that situation, and just thought was a nice way to test from A to Z (client to server) ... but we don't have to do it.

apgsga-jhe:
@chhex  : Thanks for the curl examples !!

chhex:
@apgsga-jhe  I added some curl examples, basically

1.  A curl command, which retrieves the Jenkins Crumb , which may be needed for some Rest API calls
2. A curl command , which retrieves the Data for a Job as Json and from it retrieves the lastBuildNumber

Concerning the Input Action Steps. I would split that into a separate Task and implement it, when needed. I can do that for example, since it is need for the Build. 

Basically one needs to implement the Pipeline Rest API /wfapi/pendingInputActions call via Curl , see https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/master/rest-api/README.md 
And from the json output parse for "proceedUrl" resp "abortUrl" 
And post these respectively via curl.

PR #106, opened 2020-09-01
IT-35951 / Patchfile as Parameter, remove com.offbytwo.jenkins dependency

apgsga-jhe:
For sure not perfect yet, but probably ready for a first review.

As already discussed, the whole integration is difficult to test. I wrote couple of tests within "apg-patch-service-core/src/test/java/com/apgsga/microservice/patch/core/commands/jenkins/ssh/test", but this was mainly to verify that the basics are working. All tests are ignored by default because too many things on a running Jenkins server are required. 

In terms of testing .... well .... it sounds a bit wild, but I believe the best test will be when we'll start to integrate with our Pipelines. I believe we'll only at this point realize what could miss (as couple of things will be done differently in new pipelines), or where a parameter should be passed differently, etc .... I would then create specifics ticket to be quickly implemented, when well need ...

However, having said that, I believe this is ready for a first sets of feedback about the general structure, and about the approach taken with a mix of jenkinsCli and curl calls.

Also, even if not perfect yet, best tests for me would be that we embed it within our Test-VM, and start using it ...

Not sure if the easiest is to check the differences on pull request, or to take the branch in intelliJ ;)


apgsga-jhe:
@chhex : As discussed, I modified couple of things on my branch:

- https://github.com/apgsga-it/piper/pull/106/commits/fd5001e4ccd699617985f9029f5eb48707109947
- https://github.com/apgsga-it/piper/pull/106/commits/38cce068b41ae245dc728629dcf7d19cca883f7b
- https://github.com/apgsga-it/piper/pull/106/commits/3297166bd04c77ba2af6c3383ce449cd72bf14f6
- https://github.com/apgsga-it/piper/pull/106/commits/826ee56af4b12cdb0c77dba6e9d394f104a37432
- https://github.com/apgsga-it/piper/pull/106/commits/2c1f2952d6318f92bad6af4dc6e522c9152e8904
- https://github.com/apgsga-it/piper/pull/106/commits/e218130180921644f62b34aeac3d9cfe42760cf8

Tests ran fine from both IntelliJ and from command line ...

I'll now merge the pull request.

PR #110, opened 2020-09-28
IT-36433 / apscli, start a Jenkins Job without DB integration

apgsga-jhe:
Probably we'll need more of these "detached-DB" tasks, but for now, here the first one we'll need in order to start a Jenkins job from apscli.
We can start it either with or without Jenkins.job parameter. If we provide parameter, then it will be with following format:

- p1@=v1@:p2@=v2@:p3@=v3

The reason for this format is to have kind of unique delimiter, which shouldn't interfer with a posible JSON string passed as parameter.

I also took the opportunity to clean up couple of things. 

Ideally, I would like to merge it tomorrow (29.09.2020).


apg-patch-service-cmdclient/pkg/rpm/bin/apscli.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-java -Dspring.profiles.active=default -Dlogback.statusListenerClass=ch.qos.logback.core.status.NopStatusListener -DappPropertiesFile=file:/etc/opt/apg-patch-cli/application.properties -DopsPropertiesFile=file:/etc/opt/apg-patch-cli/ops.properties -jar /opt/apg-patch-cli/bin/apg-patch-cli-fat.jar $@
\ No newline at end of file
+java -Dlogback.statusListenerClass=ch.qos.logback.core.status.NopStatusListener -jar /opt/apg-patch-cli/bin/apg-patch-cli-fat.jar $@
```
chhex, line 2: @apgsga-jhe  i don’t think it should be necessary to use the Nop Logger anymore....  if yes, then something is wrong with the contract with the caller , eg Pipeline. Stdout is simply from my perspective a too fragil contract. 
On the other hand the logging should be probably defined with a reasonable default from within the jar.

apg-patch-service-server/packaging/conf/application.properties
```diff
@@ -5,7 +5,6 @@ server.port=9010
 endpoints.shutdown.enabled=true
 management.security.enabled=false
 logging.file=log/apg-patch-service-server.log
```
chhex, line 7: @apgsga-jhe hmmmm, i would get rid of the windows newline ...

chhex (approved):
@apgsga-jhe Looks straigh forward. Just a couple of smallish remarks below

apg-patch-service-cmdclient/pkg/rpm/bin/apscli.sh
apgsga-uge: @chhex - probably partly off-topic here, but anyway, I'm trying to get a little more active in github ;-)...
I agree about stdout (as most of logging anyway) lacking the necessary formalisation for proper examination of success of a process but nevertheless it's sometimes the best or even only way to get information (probably not necessarily in this context)

apg-patch-service-server/packaging/conf/application.properties
apgsga-uge: @chhex : just a question about these special characters... they always let me "panic" a little bit ;-) what's your experience about them? could they cause errors (perhaps even by misinterpretation only) from whithin java properties files?

apgsga-jhe: @chhex , @apgsga-uge : FYI, in order to remove them, I made a "dos2unix" on the file before pushing it back to Github.

PR #112, opened 2020-10-05
IT-36392 / stageMapping saved within Patch object

apgsga-jhe:
@chhex, @apgsga-uge  : I'll probably merge this one quickly. I believe the important point to noticed is within SimplePatchContainerBean, where we save the info only when the patch does not exist. That means, we cannot change the stageMappings for an existing Patch, which I believe makes sense ... but yep, we could discuss it.


apgsga-jhe:
@chhex : I'll now merge this one, but won't delete it in case we need to check something.

PR #115, opened 2020-10-13
application.properties as Puppet template 

chhex:
@apgsga-jhe  @apgsga-uge 

What do think of the this pull request? 
The basic idea is that the initial rpm Install is executed from a puppet bolt plan, where in second step the application.properties is treated as puppet template, where the effective values are taken from the inventory.yaml file of the plans.  See also : https://github.com/apgsga-it/patchserver-setup/pull/2/files
Note: The plans have not yet been adopted


apgsga-jhe:
@chhex : Does that mean that a Piper installation could be correctly made (I mean where variable are getting substituted with values) only with Puppet? Or other said, that when installing directly with yum/rpm (without Puppet), one has to first replace the variable name before starting the service?
That would be fine for me, but if my understanding is correct, I would then maybe configure the application.properties to be also a "NO REPLACE" file.
@apgsga-uge

chhex:
@apgsga-jhe No, as discussed: The properties values substitution is only for the initial setup, respectively the first installation on a platform. After the normal rpm update mechanism takes place. 
The ratio for the initial setup scenario is that we have different "profiles" of setup targets (Local Vm, Integration Test, Production), which has distinct initially set values for some properties. 
@apgsga-uge

chhex:
@apgsga-jhe @apgsga-uge 
Here a reviewable version of the Patch Server Service Management with Bolt Puppet, see also this pull request : https://github.com/apgsga-it/patchserver-setup/pull/3/files 

The process: 
1.  The apg-patch-server-service rpm installs the property file ops.properties and application.properties as Puppet epp template files : ops_properties.initial and application_properties.intial 
2. The rpm post process does not start the service automatically
3. The puppet module piper_properties.pp reads downloads both files and processes them as template, with the values taken from the inventory.yaml file , see https://github.com/apgsga-it/patchserver-setup/pull/3/files and starts the service.

To note: 

- The piper_properties.pp  is implemented with the intention to be idempotent.  TO BE REVIEWED
- Currently the plan is implemented to be able to update to properties files, after taking a backup. TO BE DISCUSSED

Open: 
- Our requirements in terms of permissions

apgsga-jhe:
@chhex , @apgsga-uge 

Looking good to me

chhex:
@apgsga-jhe @apgsga-uge  will merge , see also https://github.com/apgsga-it/patchserver-setup/pull/3

PR #116, opened 2020-11-10
IT-36501 and IT-36392 / Piper should run without DB Integration

apgsga-jhe:
This is probably not complete yet, but enough to be merged in order to continue with other tickets without coming into merge conflicts ... Basically, in this pull request:

- New property to tell Piper we run without DB
- Patch getting all info from ServiceMetadata, since we need almost everything.
- Persisting some more properties into Service API
- Setting service packager name when pre-saving a patch
- jenkins port as specific property to differentiate between ssh and curl URL (no need to have port number for SSH)
- some additional logging

... probably some points I forgot to mention.


apgsga-jhe:
@chhex : Not sure if you had the notification regarding this one, ideally I would like to merge before going forward, ok for you ?

chhex:
@apgsga-jhe  Fine for me.
Off tone : The Service Meta Data is likely to change end to end... with mulitservice_cm , but not yet really sure in which direction. I think there is a ticket.

apgsga-jhe:
> @apgsga-jhe Fine for me.
> Off tone : The Service Meta Data is likely to change end to end... with mulitservice_cm , but not yet really sure in which direction. I think there is a ticket.

@chhex : Thanks. Yep sure regarding the serviceMetaData, pretty sure it will evolves ... Ok, I'm merging now.

PR #117, opened 2020-11-10
IT-36392 / Multiservice Build Pipeline

apgsga-jhe:
@chhex , @apgsga-uge : This one is ready to be reviewed.

As you'll notice, I removed couple of infos from Patch API. Because the Build Pipeline will be split into several Pipelines, I don't believe we need to store these info in the Patch anymore. I might realize that they'll be missing when doing the assembleAndDeply and Install Pipelines, but I don't believe ...


apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/Service.java
```diff
@@ -32,7 +32,9 @@ public static Service create() {
     private String baseVersionNumber;
     private String revisionMnemoPart;
     private boolean installOnEmptyModules = false;
+    //TODO JHE (16.11.2020): Not sure revision will be needed anymore
```
apgsga-uge, line 35: as I understood, revision should be defined per target environment and therefore shouldn't be necessary within the patch-config - I'd agree with you...

apg-patch-service-cmdclient/src/main/groovy/com/apgsga/patch/service/client/PatchCli.groovy
```diff
@@ -94,7 +95,7 @@ class PatchCli {
 			// TODO (CHE,13.9) Get rid of the component parameter, needs to be coordinated with current Patch System (PatchOMat)
 			sta longOpt: 'stateChange', args:3, valueSeparator: ",", argName: 'patchNumber,toState,component', 'Notfiy State Change for a Patch with <patchNumber> to <toState> to a <component> , where <component> can only be aps ', required: false
```
apgsga-jhe, line 96: As discussed with @chhex , "sta" option will be replace with a "build" option to which will give only the following parameters: patchNumber,target
@apgsga-uge : FYI

chhex (approved):
@apgsga-jhe  Looks good to me. Significantly less and simpler code = left side more red , then green on the right side ;-) , what is to be expected and desired. 
I don't think that revision in Patch is necessary any more,  since Revision is managed else where in and with the Gradle Plugin. 
Also i would not use the Patch Object as "state" Container for the Pipelines in any case anymore, but create a specific and simpler Groovy Object via Pipeline Library, which can be stashed und unstashed to manage Pipeline State.

PR #119, opened 2020-12-03
Service Metadata Refactoring

chhex:
API Refactoring to Reflect needed ServiceMetaData. 
Decisions: 

- Value Objects on the API , where possible and with reasonable effort
- For the Object on the API use Lombok and Jackson


chhex:
@apgsga-jhe  This is definitely work in progress. But if you want to a have a first look where the api is heading. 

apg-patch-artifact-querymanager  has been adopted to use the new API Objects and Tests are running. 
Also there are Tests in https://github.com/apgsga-it/piper/blob/che-IT-36505/apg-patch-service-api/src/test/groovy/com/apgsga/microservice/patch/api/PatchDataTests.groovy and https://github.com/apgsga-it/piper/blob/che-IT-36505/apg-patch-service-api/src/test/groovy/com/apgsga/microservice/patch/api/PatchServiceServiceMetaDataTests.groovy. which reflect the changes.

Also: I moved the Exception Handling Classes from the api back to the core code. 
Result: apg-patch-artifact-querymanager throws it's own Runtime Exceptions

apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/PatchBuildPipelineParameter.java
```diff
@@ -0,0 +1,5 @@
+package com.apgsga.microservice.patch.api;
+// TODO (CHE, Build Pipeline "View" of the Data , definitely not here
+public class PatchBuildPipelineParameter {
```
apgsga-jhe, line 3: @chhex : Not sure of the purpose of this class. Is this not redundant of "apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/BuildParameter.java" ?

chhex: @apgsga-jhe Sorry, late .... ideally the Data , in the end the Json object resp. Objects which are passed to Jenkins, should be assemble in the Implementing Class , JenkinsClientImpl.
But this work in progress. I don’t think the Objects sent as Parameter to the Pipelines correspond 1:1 to API Objects

apgsga-jhe: @chhex : Ah ok ok, I get it. Then the name I choose for BuildParameter class was bad, because it was actually meant to be passed to the Pipeline. 
"_I don’t think the Objects sent as Parameter to the Pipelines correspond 1:1 to API Objects_": Yep agreed, not necessarily, except if we create object specific for pipelines (Build Parameter for example).

PR #126, opened 2021-01-12
IT-36783 / Branch is attempted even though DBPatchBranch is null

apgsga-jhe:
When first saving a Patch, DbPatchBRanch and prodBranch should be set. Default values are provided, but could be customized with following Piper's properties:

- db.patch.branch.prefix
- db.prod.branch


chhex (changes requested):
@apgsga-jhe  I would consider to move the 

```
    String dbPatchBranch;
    @Builder.Default
    String prodBranch = PROD_BRANCH_DEFAULT;
```

into DbPatch and there then use the with Pattern as i used in MavenArtifact, see https://github.com/apgsga-it/piper/blob/master/apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/MavenArtifact.java
It makes the "deviation" of the Value Object Pattern , clear , on a attribute level. 

If you don't have time, i can do this..... the effort should be minimal
So my request changes is more a propose changes. 

But in terms of separations of concerns , i - personally - would go for it

apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/Patch.java
```diff
@@ -15,12 +15,9 @@
 @Builder(toBuilder = true)
 public class Patch {
 
-    private static final String PROD_BRANCH_DEFAULT = "prod";
-
     String patchNumber;
     String dbPatchBranch;
```
chhex, line 19: @apgsga-jhe As said , i would move these two to the DbObjects

chhex (commented):
@apgsga-jhe  So the Json would be instead of

```
{
  "patchNumber": "1",
  "dbPatchBranch": null,
  "prodBranch": "prod",
  "patchTag": "",
  "developerBranch": "",
  "tagNr": 0,
  "dbObjects": [],
  "dockerServices": [],
  "services": [
    {
      "serviceName": "testapp-service",
      "artifactsToPatch": [
        {
          "artifactId": "testapp-module",
          "groupId": "com.apgsga.testapp",
          "name": null,
          "version": "1.0.0.DEV-ADMIN-UIMIG-SNAPSHOT",
          "dependencyLevel": 0
        }
      ],
      "serviceMetaData": null
    }
  ]
}
```

Something like"

{
  "patchNumber": "1",
  "patchTag": "",
  "developerBranch": "",
  "tagNr": 0,
  "dbPatch": {
       "dbPatchBranch": null,
       "prodBranch": "prod",
       "dbObjbects" []
   },
  "dockerServices": [],
  "services": [
    {
      "serviceName": "testapp-service",
      "artifactsToPatch": [
        {
          "artifactId": "testapp-module",
          "groupId": "com.apgsga.testapp",
          "name": null,
          "version": "1.0.0.DEV-ADMIN-UIMIG-SNAPSHOT",
          "dependencyLevel": 0
        }
      ],
      "serviceMetaData": null
    }
  ]
}

apgsga-jhe:
@chhex : See also my comment on the ticket: https://jira.apgsga.ch/browse/IT-36783?focusedCommentId=361108&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-361108

Agreed in terms of implementation, I'll adapt accordingly, and re-submit for Review.

apgsga-jhe:
@chhex : As discussed on the phone, I'll merge (unused with* methods have been removed).

PR #127, opened 2021-01-13
Refactored PatchTag

chhex:
@apgsga-jhe  As discussed

I moved the PatchTag field to DbPatch resp Service. 
To minimize the Change for BuildPipelineParameter, i only introduced one new field: 

`    String dbPatchTag;`

To consider : Use DbPatch Object in BuildPipelineParameter instead of 

```
    List<String> dbObjectsAsVcsPath;
    String dbPatchBranch;
    String dbPatchTag;
    List<DbObject> dbObjects;
```

I also refactored Patch and  Service to use the with Pattern for updatable attributes. 
And i removed some unused (old) testdata. 

TODO : Adopt Pipeline accordingly, after Review.


apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/PatchSetupTask.java
```diff
@@ -5,7 +5,6 @@
 import com.apgsga.microservice.patch.api.*;
```
chhex, line 5: @apgsga-jhe Here we are simply updateing the Patch Object Graph now, instead of creating a copied Graph.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/jenkins/TaskStartBuildPipeline.java
```diff
@@ -25,21 +25,19 @@
     private final String jenkinsSshUser;
     private final JenkinsPipelinePreprocessor preprocessor;
     private final BuildParameter buildParameters;
-    private final String dbLocation;
```
chhex, line 28: dbLocation was used ... it deleted it.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/PatchSetupTask.java
```diff
@@ -39,40 +38,40 @@ private PatchSetupTask(CommandRunner jschSession, Patch patch, PatchPersistence
...
         try {
-            Patch patchWithResDep = resolveDependencies();
-            Integer lastTagNr = patchWithResDep.getTagNr();
+            resolveDependencies();
+            Integer lastTagNr = patch.getTagNr();
             Integer nextTagNr = lastTagNr + 1;
-            String patchBranch = patchWithResDep.getDbPatch().getDbPatchBranch();
-            String nextPatchTag = patchBranch + "_" + nextTagNr.toString();
-            LOGGER.info("Patch Setup Task started for patch " + patchWithResDep.getPatchNumber());
+            patch.withTagNr(nextTagNr);
```
apgsga-jhe, line 44: @chhex : I believe this could be done within the JavaService loop, similar to dbPatch.withPatchTag(nextTagNr)

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/jenkins/TaskStartBuildPipeline.java
```diff
@@ -51,9 +49,9 @@ private String pipelineBuildParameterAsJson() {
                     .stageName(buildParameters.getStageName())
                     .successNotification(buildParameters.getSuccessNotification())
                     .errorNotification(buildParameters.getErrorNotification())
-                    .patchTag(patch.getPatchTag())
```
apgsga-jhe, line 54: @chhex : As you wrote, you'll adapt the pipeline once this will be merged. But I believe you'll need a "javaServicePatchTag" to be passed as part of the parameters to the pipeline.

apgsga-jhe (changes requested):
@chhex : Just added 2 comments, but otherwise, looking fine for me.

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/jenkins/TaskStartBuildPipeline.java
chhex: @apgsga-jhe  Don't think i should , but will have to look at the Pipeline code. I assume that i can access the patchTag field transitively via Service object... naturally this requires the equivalent loop in the Pipeline code.

apgsga-jhe: @chhex : sounds good (y)

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/PatchSetupTask.java
chhex: @apgsga-jhe  Not quite sure what you mean here. I thought we said that the TagNr is per Patch and not per DbPatch resp Services? So it shouldn't be in a loop. So this code simply sets the new tagNr. 
Ok, what could be done: 
` Integer nextTagNr = lastTagNr + 1;`
within the withTagNr method.
You meant that?

apgsga-jhe: @chhex : Oh sorry, yeah, you're calling withTagNr, and not withTagName .... All good, yes exactly, TagNr is per Patch. Then ... all good for me :)

apgsga-jhe (approved):

apg-patch-service-api/src/main/java/com/apgsga/microservice/patch/api/Patch.java
```diff
@@ -5,41 +5,45 @@
...
+    @Getter
     DBPatch dbPatch;
     @Builder.Default
+    @Getter
     List<String> dockerServices = Lists.newArrayList();
     @Builder.Default
+    @Getter
     List<Service> services = Lists.newArrayList();
 
+    public void nextTagNr() {
```
chhex, line 37: @apgsga-jhe  Yep, i inlined the increment Tag

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/PatchSetupTask.java
chhex: @apgsga-jhe  I inlined the Tagnr incrementation, so now there is there is a nextTagNr function called.

apgsga-jhe: > @apgsga-jhe I inlined the Tagnr incrementation, so now there is there is a nextTagNr function called.

@chhex  : Perfect, cheers!

PR #134, opened 2021-02-10
IT-35965 / Clone pipeline

apgsga-jhe:
Provide possibility to start an onClone pipeline with specific parameters.

### DIFF
--- a/apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/SimplePatchContainerBean.java
+++ b/apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/SimplePatchContainerBean.java
@@ -346,6 +346,25 @@ public class SimplePatchContainerBean implements PatchService, PatchOpService {
 	}
 
 	@Override
+	public void startOnClonePipeline(OnCloneParameters parameters) {
+		LOGGER.info("Starting onClone pipeline with following parameter: " + parameters.toString());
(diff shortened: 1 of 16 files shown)


apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/jenkins/OnDemandPipelineParameter.java
```diff
@@ -16,7 +16,6 @@
 
     String patchNumber;
     String target;
-    String patchTag;
```
apgsga-jhe, line 19: This one was actually unused

apg-patch-service-core/src/main/java/com/apgsga/microservice/patch/core/impl/SimplePatchContainerBean.java
```diff
@@ -345,6 +345,25 @@ public void startOnDemandPipeline(OnDemandParameter parameters) {
 		jenkinsClient.startOnDemandPipeline(parameters);
 	}
 
+	@Override
```
chhex, line 348: @apgsga-jhe  Small one... also not quite sure, not seeing all the code: 
In terms of encapsulation and responsibilities, would'nt it be better to move the validation and getting the ProductionTargetName to the jenkinsClient.startOnClonePipeline..

chhex (approved):
@apgsga-jhe Looking good to me

